PR #686, opened 2022-10-01
Removed the duplicate rpk container page

rkruze:
It now redirects to the macOS guide which uses rpk container. Outside of macOS `rpk container` is not normally used. 

Also updated the versioned docs to point to the right version of the previous `rpk container` page.


Feediver1:
@Deflaimun Doesn't this need a redirect for the removal of the Starting a Local Cluster topic? If so, can you add that here? I suspect that @rkruze is not familiar with the redirect requirements. Thanks.

rkruze:
@Feediver1 I updated the `_redirect` file which seems to work on the preview link, is more needed on top of that? 

You can try it here: `https://deploy-preview-686--redpanda-documentation.netlify.app/docs/deployment/guide-rpk-container`

Deflaimun:
Everything looks good, except for the link changes. They're already covered in https://github.com/redpanda-data/documentation/pull/687

Deflaimun:
I have fixed the PR. It can be merged now.

Feediver1 (approved):
Thanks @rkruze and @Deflaimun

PR #691, opened 2022-10-03
docs: new redpanda operator section

micheleRP:
This moves all the original pages with info about the Redpanda Operator into a new Reference section. 

Preview: 
https://deploy-preview-691--redpanda-documentation.netlify.app/docs/reference/redpanda-operator/
https://deploy-preview-691--redpanda-documentation.netlify.app/docs/reference/redpanda-operator/kubernetes-qs-local-access/
https://deploy-preview-691--redpanda-documentation.netlify.app/docs/reference/redpanda-operator/kubernetes-qs-minikube/
https://deploy-preview-691--redpanda-documentation.netlify.app/docs/reference/redpanda-operator/kubernetes-qs-cloud/
https://deploy-preview-691--redpanda-documentation.netlify.app/docs/quickstart/kubernetes-qs-dev/
https://deploy-preview-691--redpanda-documentation.netlify.app/docs/deployment/
https://deploy-preview-691--redpanda-documentation.netlify.app/docs/security/

PR #693, opened 2022-10-04
docs: new k8s redirects

micheleRP:
This adds new redirects to all the k8s pages that were moved under the Reference section.


micheleRP:
@Deflaimun This PR is on dev, but the preview isn't showing the changes from #691 that are live (so I can't check that these redirects work). Can you please look at it?

PR #708, opened 2022-10-12
Update doc to remove duplicate key

corymacd:
Duplicate key causes the example to throw duplicate key error.


weeco (approved):

Feediver1 (approved):
Thanks for the update @corymacd !

PR #740, opened 2022-10-23
engine: Fix path to the custom dark mode icon

JakeSCahill:
I was looking at the docs in dark mode and noticed that the icon is hard to see.

Looking at the source for the docs, you had already swizzled the `IconDarkMode` component and changed the color to `white`. But it looks like the latest `preset-classic` theme has changed the path to `Icon/DarkMode`. So I just updated the path.

Result:

image


AnneLeeperRP:
Thanks, Jake! This is great! @Deflaimun , can you please take a look and approve?

Deflaimun:
This is a good change. Thanks for noticing that @JakeSCahill !

Deflaimun (approved):

PR #750, opened 2022-10-27
engine: add Feedback UI

Deflaimun:
Resolves #639

### DIFF
--- a/src/theme/DocItem/Layout/index.js
+++ b/src/theme/DocItem/Layout/index.js
@@ -76,7 +76,7 @@ const MyModal = (props) => {
               </div>
               <div className={styles.modalFooter}>
                 {!feedbackSubmited &&
-                <button onClick={() => {setfeedbackSubmited(true); setTimeout(props.onClose,30000)}} className={clsx("button", styles.submitButton)} disabled={!disableButton}>Submit</button>
+                <button type= "submit" onClick={() => {setfeedbackSubmited(true); setTimeout(props.onClose,30000)}} className={clsx("button", styles.submitButton)} disabled={!disableButton}>Submit</button>
                 }
                 <button onClick={props.onClose}  className={clsx("button", styles.closeButton)}>Close</button>
               </div>
--- a/src/theme/DocItem/Layout/index.js
+++ b/src/theme/DocItem/Layout/index.js
@@ -50,7 +50,7 @@ const MyModal = (props) => {
                 <h4>{title}</h4>
               </div>
               <div className={styles.modalBody}>
-                <form>
+                <form data-netlify="true">
                   <div className={styles.radioButtons}>
                     <input type="radio" name="thumb-up" id="easyToUnderstand" onChange={() => {setOther(false); setDisableButton(true);}} />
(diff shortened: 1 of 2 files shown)


Deflaimun:
Not ready yet. Sending events to google tag manager is not working yet.

Feediver1:
Let's get this restarted @Deflaimun

src/theme/DocItem/Layout/index.js
```diff
@@ -1,31 +1,123 @@
...
+  otherRadio = "Other";
+
+  if(props.positiveFeedback){
+    title = "What do you like about this doc?";
+    easyRadio = "Easy to understand";
+    solvedRadio = "Solved my problem";
+    whatWeDo = "Let us know what we do well:";
+  }
+  else{
+    title = "What did you not like about this doc?";
```
JakeSCahill, line 37:
```suggestion
    title = "What don't you like about this doc?";
```

The first question uses the present tense, so just being consistent.

src/theme/DocItem/Layout/index.js
```diff
@@ -1,31 +1,123 @@
...
+    easyRadio = "Hard to understand";
+    solvedRadio = "Did not solve my problem";
+    whatWeDo = "Let us know what we can improve:";
+    
+  }
+
+  return (
+    <form data-netlify="true" netlify name="submissionForm" >
```
JakeSCahill, line 45: should we use a honeypot or something to catch spam? https://docs.netlify.com/forms/spam-filters/#honeypot-field

JakeSCahill: I'm also just genuinely curious how this gets reported in Netlify and whether we will all have access to the results.

Deflaimun:
@Feediver1 I need an approval to merge this so we can test this live.

src/theme/DocItem/Layout/index.js
Deflaimun: We kinda wrote this on purpose to make the NOT more preeminent. But we can talk about this in our meetings.

src/theme/DocItem/Layout/index.js
Deflaimun: Using honeypot to catch spam is something that we want to do too. We haven't implemented yet because either the feature is not working, or I doesn't work on previews. It's not very clear why it isn't working. We're trying to test this live to see if something changes. 

We will def have access to everything. And if I understood their docs correctly, we can do almost anything with that data.

Feediver1: @Deflaimun 
Let's get this implemented and then circle back with remaining updates (like this), as well as the form/text submission issue that was holding this up (hopefully, Docusaurus will support, or we will find a way to implement).

Feediver1 (approved):

src/theme/DocItem/Layout/index.js
Feediver1: @Deflaimun Please file a new issue after we merge this to ensure all the requested updates are fully implemented.

Issue #866, opened 2022-11-28
Update Redpanda intro about streaming and Redpanda basics


micheleRP:
several [old doc issues](https://github.com/redpanda-data/documentation/issues/151) asked for updates to the Redpanda intro

Issue #151, opened 2022-02-04
1Q 2023: Add Redpanda Introduction, Architecture, and Streaming 101 overview


Deflaimun:
**Acceptance Criteria**
Add a page: 

- Streaming 101 and basic use cases
- Kafka 101 and basic terminology (could be rolled into the above)

Feediver1:
Closing as all sub-issues have been delivered.

PR #890, opened 2022-11-30
docs:#865 sizing guidelines and use cases

### LINKED ISSUE #865: Architecture and sizing
New doc page based on James Kinley's [architecture and sizing guide](https://go.redpanda.com/redpanda-sizing-guidelines).

micheleRP:
resolves #865 

This PR covers the sizing content in issue https://github.com/redpanda-data/documentation/issues/865. (The architecture content that initially started in the PR moved to https://github.com/redpanda-data/documentation/pull/866.)

This adds a page about sizing guidelines and a page with sizing use cases (based on https://go.redpanda.com/redpanda-sizing-guidelines). 

See preview docs:
- https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing/
- https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing-use-cases/
- https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/


docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+This is encapsulated in the Kafka API, so clients can continue to produce and consume data from Redpanda in the same way. Consumers that keep up with producers continue to read from local storage and are subject to the local data retention policy. Consumers that want to read from older offsets do so with the same consumer API, and Redpanda handles fetching the necessary log segments from object storage. 
+
+See also: [Tiered Storage](../../platform/data-management/tiered-storage/)
+
+### Production settings
+
+Before running performance benchmark testing, set Redpanda into production mode and run the auto tuning tool on every node. This enables the necessary hardware optimizations and ensures that the kernel parameters are set correctly.
```
micheleRP, line 134: SMEs: What is the auto-tuning too? Is it `rpk iotune`?

micheleRP:
@jrkinley, @dcoded: Please review this doc based on James's Architecture and Sizing paper. The overview and architecture content is a new page in the Introduction section. The sizing content is in 2 pages in the Deployment section. I'd like to merge this into our doc by Friday, Dec 9th!
Denis: I added you as a PM reviewer for this new content. If someone else from Product should do this instead, please reassign it! 
James: I didn't include the images from your paper. If you want them included, can you please send me the original PNG files?

micheleRP:
@Feediver1 A reminder that this PR just covers the architecture & sizing info from James's paper. The clean-up of all overview content will go next with https://github.com/redpanda-data/documentation/issues/151 (e.g., the other content in the [intro bucket](https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/introduction/) will change).

micheleRP:
@jrkinley, @dcoded: Reminder to please review/approve this PR. I expect with issue #151 (hopefully next sprint), the architecture content will move a bit, with the Redpanda features moving into a new introduction and this page containing just Redpanda architecture. @jrkinley please let me know if anything from the existing [Designed for Performance page](https://docs.redpanda.com/docs/platform/introduction/autotune/) should be integrated into this new [Sizing Guidelines page](https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing/), or if that content should be removed. Thank you!

Feediver1:
I'm curious why sizing use cases are not a subtopic of sizing guidelines? Seems weird to have them at the same heading level, as they both fall under the heading: Sizing...Maybe have:
Sizing
  Guidelines
  Use Cases
Thoughts?

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+## Sizing considerations
+
+### Network
+
+A basic Redpanda cluster can have three nodes, a topic with a single partition and a replication factor of three, and a single producer and consumer. For every 100 MB written to the partition’s leader, 200 MB is transmitted across the network to other nodes for replication, and 100 MB is transmitted to the consumer. In this case, producing 100 MB/sec equates to 400 MB/sec (3,200 Mbps) of network traffic. Even with the same amount of data produced, increasing the replication factor or the number of consumers increases bandwidth utilization.
+
+It’s important to measure the network bandwidth between nodes, and between clients and nodes, to make sure that you’re getting the expected performance from the network. This is especially important for cloud deployments where network bandwidth is not always guaranteed. For example, AWS i3en instances only guarantee network bandwidth above a certain instance size (`i3en.6xlarge`). Below that size, network bandwidth is advertised as “up to 25 Gbps”. For this reason, soak test the network to understand how it behaves over longer periods of time.
```
Feediver1, line 32: I think AWS actually uses AWS I3en (uppercase "i")

docs/platform/deployment/sizing-use-cases.mdx
```diff
@@ -0,0 +1,106 @@
...
+- Consumer rate: 8,000 messages per second
+- Consumer throughput: 2,000 MB/sec (16,000 Mbps)
+- Data retention: 1 year
+- Failure tolerance: 2 nodes
+
+This use case has many topics, hundreds of partitions, and a high throughput. The combined producer and replication data equates to 8 Gbps of network traffic. Plus 16 Gbps for the consumers and 8 Gbps for Tiered Storage. In total, that’s at least 32 Gbps of network bandwidth required to sustain this level of throughput. Writing at 1,000 MB/sec is near the upper limit of what a single NVMe disk can sustain.
+
+At this scale, you get significant performance gains by distributing the writes over many cores and disks to better leverage Redpanda’s thread-per-core model. For example, given five nodes with 24 cores each, start with at least one partition for each core (120 partitions in total) and scale up.
+
+At this scale, Redpanda generates over three TB of writes each hour and over 80 TB each day. Local storage is going to fill up quickly, and the window of opportunity for consumers to read from local storage is going to be shorter than in the other scenarios. In this use case, Tiered Storage is essential.
```
Feediver1, line 95: Make the Tiered Storage reference a link here--especially because you are saying it is essential. Makes it quicker/easier for customer to get there.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+This is encapsulated in the Kafka API, so clients can continue to produce and consume data from Redpanda in the same way. Consumers that keep up with producers continue to read from local storage and are subject to the local data retention policy. Consumers that want to read from older offsets do so with the same consumer API, and Redpanda handles fetching the necessary log segments from object storage. 
+
+See also: [Tiered Storage](../../platform/data-management/tiered-storage/)
+
+### Production settings
+
+Before running performance benchmark testing, set Redpanda into production mode and run the auto-tuning tool on every node. This enables the necessary hardware optimizations and ensures that the kernel parameters are set correctly.
+
+See also: [Set Redpanda production mode](../../platform/deployment/production-deployment/#set-redpanda-production-mode)
```
Feediver1, line 136: This link doesn't work

Feediver1 (approved):
See the comments I shared

micheleRP:
> I'm curious why sizing use cases are not a subtopic of sizing guidelines? Seems weird to have them at the same heading level, as they both fall under the heading: Sizing...Maybe have: **Sizing**
> 
> * Guidelines
> * Use Cases
> 
> Thoughts?

@Feediver1: Yeah, I thought about that, but that would nest it 3 levels in our nav tree (Deployment/PerformanceTuning/Sizing/UseCases). Is that OK? We have rpk commands nested that deep, but no doc pages.

docs/platform/deployment/sizing.mdx
micheleRP: fixed, thanks!

docs/platform/deployment/sizing-use-cases.mdx
micheleRP: fixed

docs/platform/deployment/sizing-use-cases.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+- Average message size: 250 KB
+- Producer rate: 4,000 messages per second
+- Producer throughput: 1000 MB/sec (8,000 Mbps)
+- Consumer rate: 8,000 messages per second
+- Consumer throughput: 2,000 MB/sec (16,000 Mbps)
+- Data retention: 1 year
+- Failure tolerance: 2 nodes
+
+This use case has many topics, hundreds of partitions, and a high throughput. The combined producer and replication data equates to 8 Gbps of network traffic. Plus 16 Gbps for the consumers and 8 Gbps for Tiered Storage. In total, that’s at least 32 Gbps of network bandwidth required to sustain this level of throughput. Writing at 1,000 MB/sec is near the upper limit of what a single NVMe disk can sustain.
```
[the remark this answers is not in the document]
micheleRP: fixed

docs/platform/deployment/sizing.mdx
micheleRP: fixed, thanks!

docs/platform/deployment/sizing-use-cases.mdx
```diff
@@ -0,0 +1,106 @@
+---
+title: Sizing Use Cases
```
jrkinley, line 2: I think this information is a little out of date. I suggest we run this past @travisdowns and @tmgstevens for input on what scenarios to include based on recent benchmark / performance testing.

jrkinley: Ideally, this should cover self-hosted (bare metal and VMs) and Redpanda Cloud.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,149 @@
...
+---
+
+<head>
+    <meta name="title" content="Sizing Guidelines| Redpanda Docs"/>
+    <meta name="description" content="How to size Redpanda clusters for low, medium, and high throughput use cases in your data center and in the cloud."/>
+</head>
+
+For best performance, size your Redpanda cluster to handle the volume of data being produced, replicated, and consumed. The following variables affect cluster sizing:
+
+- MB/sec of data being produced (after compression, if applied)
```
jrkinley, line 12: - MB/sec of data being produced and consumed

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+The settings are enforced at the partition level. Each partition writes to its own data directory, into one or more log segments. The size of each log segment is determined by `segment.bytes` (default = one GB). As data is produced on a partition, it is appended to the open log segment. When that log segment reaches `segment.bytes` in size, it is closed, and a new log segment is opened.
+
+- `retention.bytes`: This controls the maximum size of a partition directory. When the total size of all log segments in a partition directory reaches this value, the retention process discards old log segments. Multiply `retention.bytes` by the number of partitions in your topic to get the topic-level retention in bytes.
+
+- `retention.ms`: This also controls the size of a partition directory, but it does so based on the age of log segments rather than their combined size. Log segments are discarded when their maximum timestamp grows older than `retention.ms`.
```
jrkinley, line 116: As above, the behaviour of `retention.bytes` and `retention.ms` has changed in `v22.3`.

jrkinley (changes requested):
Redpanda has moved on somewhat since we published the architecture and sizing guide. Specifically with "cloud first" storage in `v22.3` and the latest performance benchmarks. I think these docs need to be refreshed to include some of the latest messaging and positioning covered in: https://redpanda.com/blog/whats-new-in-redpanda-22-3.

micheleRP:
> > I'm curious why sizing use cases are not a subtopic of sizing guidelines? Seems weird to have them at the same heading level, as they both fall under the heading: Sizing...Maybe have: **Sizing**
> > 
> > * Guidelines
> > * Use Cases
> > 
> > Thoughts?
> 
> @Feediver1: Yeah, I thought about that, but that would nest it 3 levels in our nav tree (Deployment/PerformanceTuning/Sizing/UseCases). Is that OK? We have rpk commands nested that deep, but no doc pages. Do you mean to move it into a separate Sizing section at the same level as Performance and Tuning?

@Feediver1 can you please confirm: Does your thumbs up mean I should move Sizing Guidelines up to the same level as Performance and Tuning, with Sizing Use Cases nested under it? Or that I should leave the structure as it is right now?

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,143 @@
...
+time_based
+runtime=300
+[file1]
+size=10G
+ioengine=libaio
+iodepth=16
+$ sudo fio fio-seq-write.job
+```
+
+Key performance metrics:
```
micheleRP, line 100: @jrkinley should we replace this with a link to the metrics: https://docs.redpanda.com/docs/platform/cluster-administration/monitoring/?

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,143 @@
...
+
+This is encapsulated in the Kafka API, so clients can continue to produce and consume data from Redpanda in the same way. Consumers that keep up with producers continue to read from local storage and are subject to the local data retention policy. Consumers that want to read from older offsets do so with the same consumer API, and Redpanda handles fetching the necessary log segments from object storage. 
+
+See also: [Tiered Storage](../../data-management/tiered-storage/)
+
+### Production settings
+
+Before running performance benchmark testing, set Redpanda into production mode and run the auto-tuning tool (`rpk redpanda tune all`) on every node. This enables the necessary hardware optimizations and ensures that the kernel parameters are set correctly.
```
micheleRP, line 128: @jrkinley please confirm if `rpk redpanda tune all` is correct here.

docs/platform/deployment/sizing.mdx
micheleRP, line 100: @jrkinley should we replace this list of key performance metrics with a link to the latest metric doc page: https://docs.redpanda.com/docs/platform/cluster-administration/monitoring/?

micheleRP:
Status: Adding @tmgstevens and @travisdowns as reviewers for input on what scenarios to include based on recent benchmark / performance testing. Per James: ideally, this should cover self-hosted (bare metal and VMs) and Redpanda Cloud.

micheleRP:
Marking as blocked until Travis and Tristan can review.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,143 @@
...
+- Topic replication factor
+- Number of producers and consumers
+
+Throughput and retention requirements can cause bottlenecks in the system. On an undersized cluster, clients could saturate the available network bandwidth, a disk could run out of IOPS and be unable to keep up with writes, or you could simply run out of disk space. On an oversized Redpanda cluster, you could overpay for unnecessary infrastructure. For detailed sizing on various throughput and retention scenarios, see [Sizing Use Cases](../../deployment/sizing-use-cases).
+
+This topic describes sizing considerations. In general:
+- 2:1 core-to-local-disk ratio for maximum I/O performance (preferably NVMe SSD)
+- Minimum of 2GB memory for each core, but more memory is better
+- Total network bandwidth needs to account for writes, replication, and reads
+- Test with I3en (AWS), n2-standard (GCP), and lsv2 (Azure) cloud instance types
```
tmgstevens, line 22: Type: I3 should be i3

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+- Total network bandwidth needs to account for writes, replication, and reads
+- Test with i3en (AWS), n2-standard (GCP), and lsv2 (Azure) cloud instance types
+- Use Tiered Storage to unify historical and real-time data
+- Run hardware and Redpanda benchmark tests to provide a performance baseline
+
+## Sizing considerations
+
+### Network
+
+A basic Redpanda cluster can have three nodes, a topic with a single partition and a replication factor of three, and a single producer and consumer. For every 100 MB written to the partition’s leader, 200 MB is transmitted across the network to other nodes for replication, and 100 MB is transmitted to the consumer. In this case, producing 100 MB/sec equates to 400 MB/sec (3,200 Mbps) of network traffic. Even with the same amount of data produced, increasing the replication factor or the number of consumers increases bandwidth utilization.
```
shorgi, line 30:
```suggestion
To understand the network usage of a basic Redpanda cluster, consider a cluster with three nodes, a topic with a single partition and a replication factor of three, and a single producer and consumer. For every 100 MB written to the partition’s leader, 200 MB is transmitted across the network to other nodes for replication, and 100 MB is transmitted to the consumer. In this case, producing 100 MB/sec equates to 400 MB/sec (3,200 Mbps) of network traffic. Even with the same amount of data produced, increasing the replication factor or the number of consumers increases bandwidth utilization.
```

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+[ 5] 3.00-4.00 sec 1.11 GBytes 9.53 Gbits/sec 0 1.72 
+MBytes
+[ 5] 4.00-5.00 sec 1.11 GBytes 9.53 Gbits/sec 0 1.72 
+MBytes
+...
+```
+
+### CPU and memory
+
+Topic partitions are the unit of parallelisation in Redpanda. Adding partitions is how you scale to meet workload demands. Redpanda is designed to scale up to utilize all available hardware and scale out to distribute performance across multiple nodes.
```
shorgi, line 61:
```suggestion
Redpanda is designed to scale up to utilize all available hardware and scale out to distribute performance across multiple nodes. Topic partitions are the unit of parallelization in Redpanda. Adding partitions is how you scale to meet workload demands. 
```

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+Redpanda implements a thread-per-core programming model through its use of the [Seastar](https://seastar.io/) library. This allows Redpanda to pin each of its application threads to a CPU core to avoid context switching and blocking, significantly improving processing performance and efficiency.
+
+Redpanda can handle approximately one GB/sec of writes for each core, depending on the workload. Since NVMe disks can have a sustained write speed of over one GB/sec, it takes two cores to saturate a single NVMe disk. At higher throughput rates, a 2:1 core-to-disk ratio is important to avoid I/O bottlenecks, and a minimum of four cores is needed.
```
shorgi, line 65:
```suggestion
Redpanda can handle approximately one GB/sec of writes for each core, depending on the workload. Since NVMe disks can have a sustained write speed of over one GB/sec, it takes two cores to saturate a single NVMe disk. At higher throughput rates, a 2:1 core-to-disk ratio is important to avoid I/O bottlenecks, so a minimum of four cores is needed.
```

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+Redpanda is basically a distributed transaction log, with well-understood access patterns. It appends data to the end of log files, and sequentially reads data from log files. For this reason, Redpanda bypasses the Linux page cache and manages its own memory and disk I/O. This gives Redpanda complete control over the underlying hardware to optimize I/O performance, deliver predictable tail latencies, and minimize its memory footprint. A minimum of two GB of memory for each core is recommended, but more is better.
```
shorgi, line 67:
```suggestion
Redpanda is basically a distributed transaction log with well-understood access patterns. It appends data to the end of log files and sequentially reads data from log files. For this reason, Redpanda bypasses the Linux page cache and manages its own memory and disk I/O. This gives Redpanda complete control over the underlying hardware to optimize I/O performance, deliver predictable tail latencies, and minimize its memory footprint. A minimum of two GB of memory for each core is recommended, but more is better.
```

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+### Storage
+
+Your best storage solution for your workload depends on your performance and data retention requirements. 
+
+If high throughput and low latency is most important, then use locally-attached NVMe SSD disks. This is also a good option in the cloud. Just remember that local disks are ephemeral and data is wiped when an instance is restarted. An alternative in the cloud is to use SSD-backed remote storage to persist data between instance restarts. The hyper-clouds provide options for guaranteeing throughput and provisioned IOPS performance.
```
shorgi, line 73: What are `hyper-clouds`?

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+See also: [Configure message retention](../../deployment/disk-utilization/#configure-message-retention)
+
+### Tiered Storage
+
+With only local storage, data retention is limited to the provisioned capacity: you must provision more nodes to increase capacity. Adding nodes is expensive, because you’re forced to overprovision infrastructure regardless of whether you need the additional compute power. In most cases, overprovisioning leads to underutilization and higher operational costs.
+
+Redpanda Tiered Storage enables multi-tiered remote storage that archives log segments to object storage in near real time. Tiered Storage can be combined with local storage to provide infinite data retention and disaster recovery on a per-topic basis. Retention properties work the same for Tiered Storage topics and local storage topics. Data is retained in the cloud until it reaches the configured time or size limit.
```
shorgi, line 118:
```suggestion
Redpanda Tiered Storage enables multi-tiered remote storage that archives log segments to object storage in near real time.

With only local storage, data retention is limited to the provisioned capacity: you must provision more nodes to increase capacity. Adding nodes is expensive, because you’re forced to overprovision infrastructure regardless of whether you need the additional compute power. In most cases, overprovisioning leads to underutilization and higher operational costs.

Tiered Storage can be combined with local storage to provide infinite data retention and disaster recovery on a per-topic basis. Retention properties work the same for Tiered Storage topics and local storage topics. Data is retained in the cloud until it reaches the configured time or size limit. 
```

docs/platform/deployment/sizing-use-cases.mdx
```diff
@@ -0,0 +1,106 @@
...
+A topic with Tiered Storage enabled can write to faster local storage managed by local retention settings, and at the same time, it can write the data to object storage managed by different retention settings, or left to grow indefinitely. Consumers that generally keep up with producers stream from local storage, but at this velocity that window of opportunity is narrower. The object store is a backup if a consumer needs to read from an older offset.
+
+|  | Bare Metal | AWS | GCP | Azure |
+| ----------- | ----------- | ----------- | ----------- | ----------- |
+| Instance Type | - | i3en.6xlarge | n2-standard-32 | F48s_v2 |
+| Nodes | 3 | 3 | 3 | 3 |
+| Cores | 24 | 24 | 32 | 48 |
+| Memory | 192 GB | 192 GB | 128 GB | 96 GB |
+| Instance Storage | 30 TB (NVMe) | 15 TB (NVM3) | 9 TB (SSD) | 384 TB (SSD) |
+| Persistent Storage | - | - | - | 20 TB (Standard SSD) |
```
tmgstevens, line 77:
```suggestion
| Persistent Storage | - | - | - | 20 TB (Standard SSD) |
| Available Local Retention | 17 hrs | 8 hrs | 5 hrs | 9 days |
```

docs/platform/deployment/sizing-use-cases.mdx
```diff
@@ -0,0 +1,106 @@
...
+| Producer MB/sec | Consumer MB/sec | Avg. Replication Factor | Nodes | Writes per node MB/sec | Reads per node MB/sec |
+| ----------- | ----------- | ----------- | ----------- | ----------- | ----------- |
+| 500 | 1,500 | 3 | 3 | 500/3 * 3 = 500 | 1500/3 = 500 |
+| 500 | 1,500 | 3 | 5 | 500/5 * 3 = 300 | 1500/5 = 300 |
+| 500 | 1,500 | 3 | 7 | 500/7 * 3 = 215 | 1500/7 = 215 |
+| 500 | 1,500 | 5 | 7 | 500/7 * 5 = 358 | 1500/7 = 215 |
+
+The additional 500 MB/sec for consumer throughput is for Tiered Storage and the bandwidth required to archive log segments to object storage. When Tiered Storage is enabled on a topic, it essentially adds another consumer’s worth of bandwidth on the network.
+
+Producing an average of 500 MB every second equates to over 40 TB of data each day. With a retention period of one year, the cluster needs over 14 PB of storage. This is more than can be squeezed into three nodes, so you’re forced to overprovision just to handle the data retention requirements. This isn’t an issue in the cloud with high-performance persistent storage options (like io2 in AWS, Extreme PD in GCP, and Ultra SSD in Azure), but at this scale, it’s very expensive.
```
tmgstevens, line 66:
```suggestion
Producing an average of 500 MB every second equates to over 40 TB of data each day. With a retention period of one year, the cluster needs over 14 PB of storage. This is more than can be squeezed into three nodes, so you’re forced to overprovision just to handle the data retention requirements. This isn’t an issue in the cloud with high-performance persistent storage options (like io2 in AWS, Extreme PD in GCP, and Ultra SSD in Azure), but at this scale, it’s very expensive.

A consideration for balancing the available local disk is exactly how many reads can be serviced from local storage. With difference instance types or attached SSDs/NVMe we result in differing amounts of local storage and hence amount of data that is available without going back to the cloud.
```

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+Throughput and retention requirements can cause bottlenecks in the system. On an undersized cluster, clients could saturate the available network bandwidth, a disk could run out of IOPS and be unable to keep up with writes, or you could simply run out of disk space. On an oversized Redpanda cluster, you could overpay for unnecessary infrastructure. For detailed sizing on various throughput and retention scenarios, see [Sizing Use Cases](../../deployment/sizing-use-cases).
+
+This topic describes sizing considerations. In general:
+- 2:1 core-to-local-disk ratio for maximum I/O performance (preferably NVMe SSD)
+- Minimum of 2GB memory for each core, but more memory is better
+- Total network bandwidth needs to account for writes, replication, and reads
+- Test with i3en (AWS), n2-standard (GCP), and lsv2 (Azure) cloud instance types
+- Use Tiered Storage to unify historical and real-time data
+- Run hardware and Redpanda benchmark tests to provide a performance baseline
```
tmgstevens, line 24:
```suggestion
- Run hardware and Redpanda benchmark tests to provide a performance baseline
- The number of nodes should be chosen based on the above criteria and additional nodes added for fault tolerance. This ensure that the system can operate with full throughput in a degraded state.
```

docs/platform/deployment/sizing-use-cases.mdx
```diff
@@ -0,0 +1,106 @@
...
+- Consumer throughput: 2,000 MB/sec (16,000 Mbps)
+- Data retention: 1 year
+- Failure tolerance: 2 nodes
+
+This use case has many topics, hundreds of partitions, and a high throughput. The combined producer and replication data equates to 8 Gbps of network traffic, plus 16 Gbps for the consumers and 8 Gbps for Tiered Storage. In total, that’s at least 32 Gbps of network bandwidth required to sustain this level of throughput. Writing at 1,000 MB/sec is near the upper limit of what a single NVMe disk can sustain.
+
+At this scale, you get significant performance gains by distributing the writes over many cores and disks to better leverage Redpanda’s thread-per-core model. For example, given five nodes with 24 cores each, start with at least one partition for each core (120 partitions in total) and scale up.
+
+At this scale, Redpanda generates over three TB of writes each hour and over 80 TB each day. Local storage is going to fill up quickly, and the window of opportunity for consumers to read from local storage is going to be shorter than in the other scenarios. In this use case, [Tiered Storage](../../data-management/tiered-storage/) is essential.
```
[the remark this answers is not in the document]
micheleRP: @shorgi: I agree that it reads better in that instance as 3.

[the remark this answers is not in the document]
hcoyote: Somewhat related: what's our style recommendations around abbreviating bits vs bytes? Does it make sense to be explicit (e.g. TBytes, MBytes) especially given we do this in some of the network bandwidth sizing (Mbps)?

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+- Topic replication factor
+- Number of producers and consumers
+
+Throughput and retention requirements can cause bottlenecks in the system. On an undersized cluster, clients could saturate the available network bandwidth, a disk could run out of IOPS and be unable to keep up with writes, or you could simply run out of disk space. On an oversized Redpanda cluster, you could overpay for unnecessary infrastructure. For detailed sizing on various throughput and retention scenarios, see [Sizing Use Cases](../../deployment/sizing-use-cases).
+
+This topic describes sizing considerations. In general:
+- 2:1 core-to-local-disk ratio for maximum I/O performance (preferably NVMe SSD)
+- Minimum of 2GB memory for each core, but more memory is better
+- Total network bandwidth needs to account for writes, replication, and reads
+- Test with i3en (AWS), n2-standard (GCP), and lsv2 (Azure) cloud instance types
```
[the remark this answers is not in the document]
micheleRP: fixed

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,143 @@
...
+
+This is encapsulated in the Kafka API, so clients can continue to produce and consume data from Redpanda in the same way. Consumers that keep up with producers continue to read from local storage and are subject to the local data retention policy. Consumers that want to read from older offsets do so with the same consumer API, and Redpanda handles fetching the necessary log segments from object storage.
```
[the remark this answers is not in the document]
micheleRP: > 

Thanks @hcoyote. Please let me know if this section is still confusing: https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing/#tiered-storage

docs/platform/deployment/sizing-use-cases.mdx
micheleRP: Thanks @tmgstevens. I added this as: "To balance the available local disk, consider exactly how many reads can be serviced from local storage. Different instance types or attached SSDs/NVMe provide different amounts of local storage, so different amounts of available data without going back to the cloud." Please let me know if my rewrite is not clear or correct.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+If high throughput and low latency is most important, then use locally-attached NVMe SSD disks. This is also a good option in the cloud. Just remember that local disks are ephemeral and data is wiped when an instance is restarted. An alternative in the cloud is to use SSD-backed remote storage to persist data between instance restarts. The hyper-clouds provide options for guaranteeing throughput and provisioned IOPS performance.
+
+For example, AWS io2 volumes offer up to 64,000 IOPS and 1,000 MB/sec throughput with single-digit millisecond latency. This is an expensive option, so if you can trade performance for cost, then AWS gp3 volumes are a good alternative. GCP has comparable options with high-end Extreme persistent disks and the lesser SSD persistent disks. Likewise, Azure has Ultra, Premium, and Standard persistent disk options for choosing the right balance of performance versus cost.
+
+Whichever option you choose, benchmark Redpanda’s underlying storage for read and write performance, at least from an I/O perspective. FIO is a good tool for replicating Redpanda’s sequential write pattern and load. The following example shows how to run FIO on a Debian-based server:
```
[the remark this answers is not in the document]
micheleRP: added this

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+It’s important to measure the network bandwidth between nodes, and between clients and nodes, to make sure that you’re getting the expected performance from the network. This is especially important for cloud deployments where network bandwidth is not always guaranteed. For example, AWS i3en instances only guarantee network bandwidth above a certain instance size (`i3en.6xlarge`). Below that size, network bandwidth is advertised as “up to 25 Gbps.” For this reason, soak test the network to understand how it behaves over longer periods of time. 
+
+The following example uses [iPerf3](https://iperf.fr/) to test the network bandwidth between two Debian-based servers:
```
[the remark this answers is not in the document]
micheleRP: added

docs/platform/deployment/sizing.mdx
micheleRP: > What are `hyper-clouds`?

@shorgi this isn't in the text anymore. It's rewritten as:
If high throughput and low latency is most important, then use locally-attached NVMe SSD disks. This is also a good option in the cloud. Just remember that in the cloud, local disks are ephemeral, so data is wiped when an instance is restarted. An alternative in the cloud is to use SSD-backed remote storage to persist data between instance restarts. Most cloud providers have options for guaranteeing throughput and provisioned IOPS performance of network-attached storage, although network-attached storage almost always exhibits slightly higher tail latencies than direct-attached storage.

docs/platform/deployment/sizing.mdx
travisdowns: I think it's both, but there is some complication around containers: running redpanda in a container is more complicated because (a) I am not sure the `rpk` well-known-io detection logic will work inside a container or it may depend on the container networking config (see https://redpandadata.slack.com/archives/C024H6DSU2H/p1673450827074599) and (b) it is less clear how to run `rpk iotune`: you can't just run it on the host or inside the container, probably it has to be run on the host then the results written to somewhere accessible to the container, which we don't really document.

In general we don't document much about `rpk iotune` except that it exists and you should run it: probably an area for improvement?

docs/platform/deployment/sizing-use-cases.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+- Average message size: 250 KB
+- Producer rate: 300 messages per second
+- Producer throughput: 75 MB/sec (600 Mbps)
+- Consumer rate: 300 messages per second
+- Consumer throughput: 75 MB/sec (600 Mbps)
+- Data retention: 3 days
+- Failure tolerance: 1 node
+
+This use case uses the minimum number of Redpanda nodes. Despite the relatively low throughput of 150 MB/sec (producer and consumer), it’s important to calculate the expected bandwidth utilization and to use a network testing tool like iperf to verify that the bandwidth is available and sustainable. With a single topic with a replication factor of three, producing 75 MB/sec generates an additional 150 MB/sec of data transmitted over the network for replication, and it generates a further 75 MB/sec for the consumer group. That equates to an aggregate network bandwidth of 300 MB/sec, or 2.4 Gbps in networking terms.
```
travisdowns, line 26: Change `for the consumer group` to `for the consumers`.

docs/platform/deployment/sizing-use-cases.mdx
```diff
@@ -0,0 +1,106 @@
...
+
```
travisdowns, line 27: It is a bit unclear that we say "this case uses the minimum number of nodes" but doesn't mention how many that is in the table or in this paragraph.

travisdowns, line 27: I think we need to be clearer about the network bandwidth when it comes to the number of nodes. We get a figure of 300 MB/s but that's spread across the N nodes in a specific way. It's note that each node needs 300 MB/s of full-duplex bandwidth.

Maybe we can talk about what a single node needs?

Also, the 150 MB/s of bandwidth for the replicas is full duplex (each byte send by a broker is received by some other broker), but the 75 MB/s producer and consumer flows are half-duplex since the client "endpoint" in each case is outside of the cluster. Probably we should mention the intra-cluster bandwidth separately, so it would be like:

150 MB/s of intra-cluster full duplex bandwidth
75 MB/s of ingress from producers
75 MB/s of egress to consumers

So in total 225 MB of both incoming and outgoing flows>

All of those numbers would be divided by the node count (3, here) to get the per-node BW requirements, under the "fully balanced" scenarios.

Note that this also discards other sources of network traffic, protocol overhead and so on. So I think you'd want at least a 2x buffer here.

hcoyote: I spent time reasoning through this yesterday. I think the 300MB/s is accurate from the leader's perspective. And to account for failovers, you have to have the same available bandwidth capacity for each follower that has the potential to take over.

It seems to get complicated as you get more topics, partitions, and replicas.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,158 @@
...
+
+To do a full network stress test, run iperf3 for every combination of network routes. Unlike iperf2, iperf3 does not accept multiple network connections.
+
+### CPU and memory
+
+Redpanda is designed to scale up to utilize all available hardware and scale out to distribute performance across multiple nodes. Topic partitions are the unit of parallelization in Redpanda. Adding partitions is how you scale to meet workload demands. 
+
+Redpanda implements a thread-per-core programming model through its use of the [Seastar](https://seastar.io/) library. This allows Redpanda to pin each of its application threads to a CPU core to avoid context switching and blocking, significantly improving processing performance and efficiency.
+
+Redpanda can handle approximately one GB/sec of writes for each core, depending on the workload. Since NVMe disks can have a sustained write speed of over one GB/sec, it takes two cores to saturate a single NVMe disk. At higher throughput rates, a 2:1 core-to-disk ratio is important to avoid I/O bottlenecks, so a minimum of four cores is needed.
```
travisdowns, line 67: I don't think this is the right way of thinking about it: a 2:1 ratio might be right for some specific combination of disks, but "1 disk" is not a specific unit of performance: disks might reasonable range anywhere from 0.1 to 4+ GB/s in throughput, while cores have a much smaller performance range. Furthermore, the cloud blurs what even is a "disk", since physical disks may be logically partitioned up among tenants and with network storage with configurable performance characteristics making a 2:1 ratio even less meaningful.

For the disk, what matters most is IOPS and throughput. For throughput, we could do something similar to network bandwidth where we describe how to calculate the aggregate write bandwidth needed, etc.

How many cores we need is a very complicated topic, in line with the _general caveat_. Maybe we should just say something very ballpark like 100 MB/s per core for now.

docs/platform/deployment/sizing-use-cases.mdx
travisdowns: @hcoyote - the assumption is that every node is both a leader and a follower, and the above analysis assumes a sufficient number of partitions for good distribution [1]. So the analysis considers a "typical" node which is the leader for 1/Nth of the partitions in a cluster with N nodes, and a follower for 2/Nths of the partitions.

My claim is the 300 MB/s is wrong in two ways:

 - It is more like 225 MB/s full duplex. The difference is because the replication traffic is full duplex (wrt the cluster) while the producer & consumer traffic is half duplex (one of the endpoints is outside the cluster).
 - This is the _total bandwidth_ needed across all nodes so for the purposes of benchmarking e.g., you would probably want a per-node figure, so you should divide by 3: each node needs to be capable of 75 MB/s full-duplex bandwidth, not 300 MB/s.
 
Here's another way to do the reasoning.

Consider 1 second worth of 75/75 (producer/consumer) MB/s traffic, which is 75 MB in from producers and 75 out to consumers. Focusing only _one_ node and considering that this is correct also for the other two nodes but just for different bytes in that 75 MB payload:

- The 75 is split 3 days, so each node will handle 25 MB of producer (in) + 25 MB of consumer traffic (out)
- With N=3 and `r=3` every partition is replicated on every node (1 leader + 2 followers)
- So the node we focus on sends that 25 MB out to both replicas (2 x 25 MB out).
- By symmetry, it will also receive 25 MB from each replication (50 MB in)


So that's totally 75 out and 75 in. Every node is the same, so summing all nodes it's 225 in and 225 out, or 450 total if we count bandwidth in a half-duplex way. However, as above I think the "per node" value is more useful since you need to check "does this machine type provide X bandwidth" rather than caring about the sum of all bandwidth requirements across the cluster. Also, as mentioned, I think many customers will want to handle the 25 MB/s external ingress/egress differently than the 50 MB/s intra-cluster traffic since it may have totally different performance limitations depending on the network infra (and different costs).

If your analysis disagrees, can you show me where?

travisdowns: I should add that though the above assumes "evenly distributed load" the totals end up the same regardless of the distribution, but the per node values would vary a lot.

That's another reason you need some headroom: to account for the inexact distribution of load (which can be both because our own partition balancing may not be even over the topics of interest, or also because the customer or client library does not write to each partition evenly).

docs/platform/introduction/architecture.mdx
```diff
@@ -0,0 +1,85 @@
...
+
+As long as there is a timing inequality between heartbeat time, election timeout, and mean time between node failures (MTBF), then Raft can elect and maintain a steady leader and make progress. So, a leader can maintain its position as long as one of the ten heartbeat messages it sends to all of its followers every 1.5 seconds is received; otherwise, a new leader is elected.
+
+If a follower triggers an election, but the incumbent leader subsequently springs back to life and starts sending data again, then it’s too late. As part of the election process, the follower (now an election candidate) incremented the term and rejects requests from the previous term, essentially forcing a leadership change. If a cluster is experiencing wider network infrastructure problems that result in latencies above the heartbeat timeout, then back-to-back election rounds can be triggered. During this period of infrastructure, unstable Raft groups may not be able to form a quorum. This results in partitions rejecting writes, but no data loss can occur.
```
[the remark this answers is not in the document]
micheleRP: this page now is in a different branch, but I'll incorporate this feedback there!

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,158 @@
...
+The following example shows how to run fio on a Debian-based server:
+
+:::note 
+Ensure that the fio job runs against the chosen block device. By default, fio operates in the local directory.
+:::
+
+```bash
+$ sudo apt -y update; sudo apt -y install fio
+$ cd /var/lib/redpanda/data
+$ sudo vim fio-seq-write.job
```
travisdowns, line 90: I don't think `vim` makes sense here. Probably something like:

```
cat > fio-seq-write.job <<EOF
... (file contents>
EOF
```

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,158 @@
...
+---
+
+<head>
+    <meta name="title" content="Sizing Guidelines| Redpanda Docs"/>
+    <meta name="description" content="How to size Redpanda clusters for low, medium, and high throughput use cases in your data center and in the cloud."/>
+</head>
+
+For best performance, size your Redpanda cluster to handle the volume of data being produced, replicated, and consumed. The following variables affect cluster sizing:
+
+- MB/sec of data being produced and consumed (after compression, if applied)
```
travisdowns, line 12: Better to say "Throughput of data ..." I think.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,158 @@
...
+- Topic replication factor
+- Number of producers and consumers
+
+Throughput and retention requirements can cause bottlenecks in the system. On an undersized cluster, clients could saturate the available network bandwidth, a disk could run out of IOPS and be unable to keep up with writes, or you could simply run out of disk space. On an oversized Redpanda cluster, you could overpay for unnecessary infrastructure. For detailed sizing on various throughput and retention scenarios, see [Sizing Use Cases](../../deployment/sizing-use-cases).
+
+Choose the number of nodes based on the following criteria, and add nodes for fault tolerance. This ensures that the system can operate with full throughput in a degraded state.
+- Have a 2:1 core-to-local-disk ratio for maximum I/O performance (preferably NVMe SSD).
```
travisdowns, line 19: Per other comment, I think this 2:1 core-to-disk ratio is mostly meaningless.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,158 @@
...
+- Have a minimum of 2 GB memory for each core, but more memory is better.
```
travisdowns, line 20: Maybe we should say "per shard" not "per core"? Because I assume we are trying to recommend 2 GB per _shard_ but _core_ could be 1 or 2 shards depending on hyperthreading and also customer may not assign all shards to cores.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,158 @@
...
+- Total network bandwidth must account for writes, replication, and reads.
+- Test with i3en, i4i, is4gen (AWS), n2-standard (GCP), and lsv2 (Azure) cloud instance types.
```
travisdowns, line 22: This seems like an unusual order to a user? They should test on every cloud type? Or maybe I'm reading it wrong, and it is trying to say: _If_ you are on AWS, try these 3 types, and _if_ you are GCP ... ?

In that case it may need to restructured since there is only 1 option listed for GCP and Azure.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,158 @@
...
+- Use Tiered Storage to unify historical and real-time data.
+- Run hardware and Redpanda benchmark tests for a performance baseline.
+
+## Sizing considerations
+
+### Network
+
+To understand the network usage of a basic Redpanda cluster, consider a cluster with three nodes, a topic with a single partition and a replication factor of three, and a single producer and consumer. For every 100 MB written to the partition’s leader, 200 MB is transmitted across the network to other nodes for replication, and 100 MB is transmitted to the consumer. In this case, producing 100 MB/sec equates to 400 MB/sec (3,200 Mbps) of network traffic. Even with the same amount of data produced, increasing the replication factor or the number of consumers increases bandwidth utilization.
```
travisdowns, line 30: As mentioned in the other doc, the 100 + 200 + 100 isn't exactly right because the 200 part (replication) requires bandwidth to both send and receive (since both endpoints are in the cluster) while the 100 values have only 1 end in the cluster. So if we count send + recv seperately (i.e., 1 broker sending 1 byte to another counts as 1 + 1 = 2) then it's 100 + 200 * 2 + 100.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,158 @@
...
+
+It’s important to measure the network bandwidth between nodes, and between clients and nodes, to make sure that you’re getting the expected performance from the network. This is especially important for cloud deployments where network bandwidth is not always guaranteed. For example, AWS i3en instances only guarantee network bandwidth above a certain instance size (`i3en.6xlarge`). Below that size, network bandwidth is advertised as “up to 25 Gbps.” For this reason, soak test the network to understand how it behaves over longer periods of time.
```
travisdowns, line 32: On AWS network bandwidth is guaranteed also on the smaller instance sizes, but AWS just chooses to list the burst bandwidth ("up to 25 Gbps") on the main overview pages. 

See here for the values:

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/storage-optimized-instances.html

Where it shows that i3en.3xlarge has a guaranteed bandwidth of 12.5 Gbps which is half that of the 6x instance size, so exactly in line with its size.

This advice is still correct in spirit because short tests may use only the burst bandwidth and get unrealistically good results because of it, but I think this should be re-worked to just mention the way bursting works and how to avoid testing only the burst bandwidth rather than making a claim about "lack of guaranteed bandwidth".

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,158 @@
...
+[ 5] 2.00-3.00 sec 1.11 GBytes 9.53 Gbits/sec 0 1.64 
+MBytes
+[ 5] 3.00-4.00 sec 1.11 GBytes 9.53 Gbits/sec 0 1.72 
+MBytes
+[ 5] 4.00-5.00 sec 1.11 GBytes 9.53 Gbits/sec 0 1.72 
+MBytes
+...
+```
+
+To do a full network stress test, run iperf3 for every combination of network routes. Unlike iperf2, iperf3 does not accept multiple network connections.
```
travisdowns, line 59: What this is asking the user to do is not clear to me. Also, why do we mention iperf2 if we aren't using it?

micheleRP: @travisdowns: I think this was suggested by @hcoyote. I removed the last line. Please see https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing/#network and let me know if it's OK.

docs/platform/deployment/sizing.mdx
micheleRP: @travisdowns: Please review this edited section: https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing/#sizing-considerations

docs/platform/deployment/sizing.mdx
micheleRP: @travisdowns: Please see this edited section (I used the same values we used in the sizing use cases page): https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing/#network

docs/platform/deployment/sizing.mdx
micheleRP: fixed

docs/platform/deployment/sizing.mdx
micheleRP: @travisdowns: I'm not positive that I read your suggestion correctly. Please let me know if this is correct: https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing/

docs/platform/deployment/sizing.mdx
micheleRP: Hi @shorgi. Would you please help me make this edit? https://github.com/redpanda-data/documentation/pull/890#discussion_r1067431877

docs/platform/deployment/sizing.mdx
micheleRP: @travisdowns: Please confirm if this is correct: https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing/#cpu-and-memory

docs/platform/deployment/sizing-use-cases.mdx
micheleRP: Thank you for the detailed description, @travisdowns and @hcoyote--this was very helpful to me! Please review this edited section: https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing-use-cases/#low-throughput-short-retention. There's also similar content now in the Sizing Guidelines pages.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,145 @@
...
+
+For best performance, size your Redpanda cluster to handle the volume of data being produced, replicated, and consumed. The following variables affect cluster sizing:
+
+- MB/sec of data being produced and consumed (after compression, if applied)
+- Topic replication factor
+- Number of producers and consumers
+
+Throughput and retention requirements can cause bottlenecks in the system. On an undersized cluster, clients could saturate the available network bandwidth, a disk could run out of IOPS and be unable to keep up with writes, or you could simply run out of disk space. On an oversized Redpanda cluster, you could overpay for unnecessary infrastructure. For detailed sizing on various throughput and retention scenarios, see [Sizing Use Cases](../../deployment/sizing-use-cases).
+
+This topic describes sizing considerations. In general:
```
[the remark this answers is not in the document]
micheleRP: No, the bullets don't exactly align with the sections.

docs/platform/deployment/sizing-use-cases.mdx
```diff
@@ -0,0 +1,106 @@
...
+Three nodes satisfy Redpanda’s minimum deployment requirement (so Raft can form quorums) and also the single node failure tolerance. The throughput is not high enough to warrant any more than two cores and a single NVMe SSD disk. Be mindful of predicted growth, and estimate how much room the cluster has now and when it might need to scale up or scale out.
```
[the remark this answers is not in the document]
micheleRP: > Does `how much room the cluster has` refer to CPU and/or disk margin?

@travisdowns I added this suggestion to this paragraph: "Be mindful of predicted growth of CPU and/or disk usage, and estimate when the cluster might need to scale up or scale out." Please let me know if that's not correct.

docs/platform/deployment/sizing-use-cases.mdx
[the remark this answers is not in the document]
micheleRP: @tmgstevens: Could you please review the latest draft of this (which includes feedback from Travis Downs), and then confirm for me what changes you want here? Are you saying that all 3 use cases should use Tiered Storage? And that all of the 20 TB storage should change to 7 TB? See  https://deploy-preview-890--redpanda-documentation.netlify.app/docs/platform/deployment/sizing-use-cases/#low-throughput-short-retention

docs/platform/deployment/sizing-use-cases.mdx
micheleRP: @jrkinley: Travis and Tristan have reviewed this now. Could you please review this updated pull request and approve? Hoping to merge it by Friday.

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,158 @@
...
+bs=16K
+direct=1
+numjobs=4
+group_reporting
+time_based
+runtime=300 # 5 minute runtime
+[file1]
+size=10G
+ioengine=libaio
+iodepth=16
```
tmgstevens, line 105:
```suggestion
```bash
$ sudo apt -y update; sudo apt -y install fio
$ cd /var/lib/redpanda/data
$ sudo cat < fio-seq-write.job << EOF
[global]
name=fio-seq-write
filename=fio-seq-write
rw=write
bs=16K
direct=1
numjobs=4
group_reporting
time_based
runtime=300 # 5 minute runtime
[file1]
size=10G
ioengine=libaio
iodepth=16
EOF
```

docs/platform/deployment/sizing.mdx
tmgstevens: No, these are specifically referring to fio

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+- Throughput of data (after compression, if applied)
+- Topic replication factor
+- Number of producers and consumers
+
+Throughput and retention requirements can cause bottlenecks in the system. On an undersized cluster, clients could saturate the available network bandwidth, a disk could run out of IOPS and be unable to keep up with writes, or you could simply run out of disk space. On an oversized Redpanda cluster, you could overpay for unnecessary infrastructure. For sizing estimates and advice for various throughput and retention scenarios, see [Sizing Use Cases](../../deployment/sizing-use-cases).
+
+In general, choose the number of nodes based on the following criteria, and add nodes for fault tolerance. This ensures that the system can operate with full throughput in a degraded state.
+- Total network bandwidth must account for writes, replication, and reads.
+- Have a minimum of 2 GB memory for each shard, but more memory is better.
```
vuldin, line 20: At some point we probably need to have a page to point users where they can find the answer to the question: "what is a shard?" But for now maybe its best to replace shard with core to avoid confusion?

```suggestion
- Have a minimum of 2 GB memory for each core, but more memory is better.
```

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,159 @@
...
+- Use Tiered Storage to unify historical and real-time data.
+- For AWS, test with either i3en, i4i, or is4gen; for GCP, test with n2-standard (GCP); and for Azure, test with lsv2.
```
vuldin, line 22:
```suggestion
- For AWS, test with either i3en, i4i, or is4gen; for GCP, test with n2-standard (GCP); and for Azure, test with lsv2. In general, choose instance types that prioritize storage and network performance.
```

vuldin (commented):
This isn't a full review, but just a few comments.

docs/platform/deployment/sizing.mdx
micheleRP: Thank you @tmgstevens!

docs/platform/deployment/sizing-use-cases.mdx
hcoyote: Ok, that definitely makes sense in a well-balanced scenario. I guess 300MB/s would be worst-case, assuming something went horribly wrong and cluster balancing wasn't working (and everything got somehow pinned to one leader node)

docs/platform/deployment/sizing.mdx
```diff
@@ -0,0 +1,159 @@
...
+For best performance, size your Redpanda cluster to handle the volume of data being produced, replicated, and consumed. The following variables affect cluster sizing:
+
+- Throughput of data (after compression, if applied)
+- Topic replication factor
+- Number of producers and consumers
+
+Throughput and retention requirements can cause bottlenecks in the system. On an undersized cluster, clients could saturate the available network bandwidth, a disk could run out of IOPS and be unable to keep up with writes, or you could simply run out of disk space. On an oversized Redpanda cluster, you could overpay for unnecessary infrastructure. For sizing estimates and advice for various throughput and retention scenarios, see [Sizing Use Cases](../../deployment/sizing-use-cases).
+
+In general, choose the number of nodes based on the following criteria, and add nodes for fault tolerance. This ensures that the system can operate with full throughput in a degraded state.
+- Total network bandwidth must account for writes, replication, and reads.
```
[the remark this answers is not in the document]
hcoyote: typo: "max rights" should probably be "max writes"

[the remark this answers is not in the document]
micheleRP: Thanks @vuldin and @hcoyote. I changed this list item to:
* Total network bandwidth must account for maximum simultaneous writes and reads, multiplied by the replication factor.

docs/platform/deployment/sizing.mdx
hcoyote: wasn't me. not sure what that comment referred to. while iperf2 server supported multiple client connections in parallel, a single server wouldn't do the full cross-sectional test (e.g. N to N pathways). but I agree that taking it out is right.

Basically, we want the user to test for a 3-node environment:

NodeA <-> Node B
NodeB <-> Node C
NodeC <-> Node A

Might actually be good to come with a standalone ansible runbook or something to help them do this test.

docs/platform/deployment/sizing.mdx
micheleRP: Thanks @travisdowns. I added info this to issue #979.

jrkinley (approved):

docs/platform/deployment/sizing.mdx
hcoyote: btw, we can probably expand the details on iperf testing in another issue, it doesn't need to hold this up.

PR #949, opened 2022-12-16
docs: Correct changing replication.factor value command

daisukebe:
The original command fails with 

```
$ rpk topic alter-config test3 --set replication.factor 3
unable to parse --set: unable to find key=value pair in "replication.factor"
```

This should be 

```
$ rpk topic alter-config test3 --set replication.factor=3
TOPIC  STATUS
test3  OK
```


Deflaimun (approved):

micheleRP:
@mmaslankaprv: This looks like an easy approval. Are you good with us merging this?

VadimPlh (approved):

PR #1058, opened 2023-01-18
docs: fix default and description for empty_seed_starts_cluster property

### LINKED ISSUE #1049: empty_seed_starts_cluster default should be `true`
### Describe the Issue

https://docs.redpanda.com/docs/reference/node-properties/#empty_seed_starts_cluster this documents that:

>While false by default for backward compatibility, empty_seed_starts_cluster is recommended to be set to true.

This appears reversed: the default is true: https://github.com/redpanda-data/redpanda/blob/dev/src/v/config/node_config.cc#L75 and *true* maintains backwards compatible behaviour. `False` would be what we'd recommend, too.

shorgi:
Fixes #1049 .

Review deadline: Fri Jan 20.

### DIFF
--- a/docs/reference/node-properties.mdx
+++ b/docs/reference/node-properties.mdx
@@ -109,10 +109,10 @@ Flag to enable developer mode, which skips most of the checks performed at start
 
 Controls how a new cluster is formed. This property must have the same value in all nodes in a cluster.
 
-If `true`, an empty [seed_servers](#seed_servers) list will denote that this node should form a cluster. At most one node in the cluster should be configured with an empty seed_servers list. If no such configured node exists, or if configured to `false`, all nodes denoted by the seed_servers list must be identical among those nodes' configurations, and those nodes will form the initial cluster.
+If `true`, an empty [seed_servers](#seed_servers) list will denote that this node should form a cluster. At most, one node in the cluster should be configured with an empty seed_servers list. If no such configured node exists, or if configured to be `false`, all nodes denoted by the seed_servers list must be identical among those nodes' configurations, and those nodes will form the initial cluster.
 
 :::tip
-While it's set `true` by default for backward compatibility, empty_seed_starts_cluster is recommended to be set `false`.
+While it's set `true` by default for backward compatibility, Redpanda recommends that you set empty_seed_starts_cluster to `false`.
 :::
 
 **Default**: true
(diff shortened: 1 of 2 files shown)


docs/reference/node-properties.mdx
```diff
@@ -109,11 +109,13 @@ Flag to enable developer mode, which skips most of the checks performed at start
 
 Controls how a new cluster is formed. This property must have the same value in all nodes in a cluster.
 
+If `true`, an empty [seed_servers](#seed_servers) list will denote that this node should form a cluster. At most one node in the cluster should be configured with an empty seed_servers list. If no such configured node exists, or if configured to `false`, all nodes denoted by the seed_servers list must be identical among those nodes' configurations, and those nodes will form the initial cluster.
```
Feediver1, line 112:
```suggestion
If `true`, an empty [seed_servers](#seed_servers) list will denote that this node should form a cluster. At most, one node in the cluster should be configured with an empty seed_servers list. If no such configured node exists, or if configured to be `false`, all nodes denoted by the seed_servers list must be identical among those nodes' configurations, and those nodes will form the initial cluster.
```

docs/reference/node-properties.mdx
```diff
@@ -109,11 +109,13 @@ Flag to enable developer mode, which skips most of the checks performed at start
...
+
 :::tip
-While `false` by default for backward compatibility, `empty_seed_starts_cluster` is recommended to be set to `true`.
+While it's set `true` by default for backward compatibility, empty_seed_starts_cluster is recommended to be set `false`.
```
Feediver1, line 115:
```suggestion
While it's set `true` by default for backward compatibility, Redpanda recommends that you set empty_seed_starts_cluster to `false`.
```

Feediver1 (approved):
See suggested edits. Your call.

docs/reference/node-properties.mdx
```diff
@@ -109,11 +109,13 @@ Flag to enable developer mode, which skips most of the checks performed at start
 
 Controls how a new cluster is formed. This property must have the same value in all nodes in a cluster.
 
+If `true`, an empty [seed_servers](#seed_servers) list will denote that this node should form a cluster. At most, one node in the cluster should be configured with an empty seed_servers list. If no such configured node exists, or if configured to be `false`, all nodes denoted by the seed_servers list must be identical among those nodes' configurations, and those nodes will form the initial cluster.
+
 :::tip
-While `false` by default for backward compatibility, `empty_seed_starts_cluster` is recommended to be set to `true`.
+While it's set `true` by default for backward compatibility, Redpanda recommends that you set empty_seed_starts_cluster to `false`.
```
JFlath, line 115: Nitpick: Should this the prop name have backticks?

```suggestion
While it's set `true` by default for backward compatibility, Redpanda recommends that you set `empty_seed_starts_cluster` to `false`.
```

JFlath (approved):

PR #1141, opened 2023-02-02
docs: Update configuration for rack awareness

JakeSCahill:
A correction for the `--set` options. The `serviceAccount` and `rbac` configs are top-level, not nested in `rackAwareness`.


Deflaimun (approved):
LGTM

PR #1149, opened 2023-02-03
engine: Hide feedback in Toc at the bottom of the page

JakeSCahill:
When users reach the bottom of the page on desktop, we hide the feedback thumbs in the ToC to avoid displaying two copies of them: one in the footer, and one in the ToC.


Deflaimun (approved):
LGTM

src/theme/DocItem/Layout/index.js
```diff
@@ -195,6 +195,22 @@ export default function DocItemLayout({ children }) {
   const { editUrl } = metadata;
   const [show, setShow] = useState(false);
   const [positiveFeedback, setPositiveFeedback] = useState(true);
+  // Hide the feedback thumbs in the Toc when the user reaches the bottom of the page.
+  useEffect(() => {
+    document.addEventListener('scroll', function(e){
+      let documentHeight = document.body.scrollHeight;
+      let currentScroll = window.scrollY + window.innerHeight;
+      // Wait until the user is [modifier]px from the bottom.
+      let modifier = 300;
```
shorgi, line 204: Future improvement: is 300 the height of the footer? Can we programmatically get the height of the footer?

shorgi (approved):
Nice! LGTM

src/theme/DocItem/Layout/index.js
JakeSCahill: I thought about this, but couldn't think of an easy answer and this seemed to work quite well. We should certainly update this in future when we have a better solution 😄

Deflaimun: Good catch. Depending on the screen size, both appear
image
Tested screen size 1200x1100

PR #1216, opened 2023-02-13
Fix typo in fio example heredoc

zleinweber:
The FIO example in the sizing guide uses a heredoc that doesn't work as expected (using Ubuntu 20.04 / Bash 5.0.17):

```
root@3301daad34ae:/# sudo cat < fio-seq-write.job << EOF
> [global]
> name=fio-seq-write
> filename=fio-seq-write
> rw=write
> bs=16K
> direct=1
> numjobs=4
> group_reporting
> time_based
> runtime=300 # 5 minute runtime
> [file1]
> size=10G
> ioengine=libaio
> iodepth=16
> EOF
bash: fio-seq-write.job: No such file or directory
root@3301daad34ae:/# cat fio-seq-write.job
cat: fio-seq-write.job: No such file or directory
```

Changing the redirection to `sudo cat > fio-seq-write.job << EOF` solves the problem:

```
root@3301daad34ae:/# sudo cat > fio-seq-write.job << EOF
> [global]
> name=fio-seq-write
> filename=fio-seq-write
> rw=write
> bs=16K
> direct=1
> numjobs=4
> group_reporting
> time_based
> runtime=300 # 5 minute runtime
> [file1]
> size=10G
> ioengine=libaio
> iodepth=16
> EOF
root@3301daad34ae:/# cat fio-seq-write.job
[global]
name=fio-seq-write
filename=fio-seq-write
rw=write
bs=16K
direct=1
numjobs=4
group_reporting
time_based
runtime=300 # 5 minute runtime
[file1]
size=10G
ioengine=libaio
iodepth=16
```

No big deal. But, figured I'd send the correction since I noticed it.


shorgi (approved):
Thanks @zleinweber!

PR #1237, opened 2023-02-15
engine: font test

### LINKED ISSUE #1232: Inconsistent rendering of RP product documentation font weight/color
### Describe the Issue

Several internal users have reported that the product doc text is difficult to see. Some see it as a lighter gray, others see that the font weight is thin/difficult to see. Users on Windows and macOS report seeing the font as black or gray. It seems to vary by OS and browser. One user reports that it is very light gray while viewing in Brave (vs Safari). Following image was shared for the Brave browser--both the black and red font render super light:
image (9)

Please ensure our CSS settings are correct, that our font color and weight are consistent between the navtrees and center text font.

@yougotashovel Can you identify why this occuring? Thx.

Deflaimun:
Resolves #1232


JakeSCahill:
@Deflaimun it seems the issue is to do with Chrome loading different fonts.

If I open the docs in two separate Chrome profiles, one looks fine and the other displays the issue.

shorgi:
@Deflaimun - try different font smoothing settings, such as `-webkit-font-smoothing: subpixel-antialiased;`

JakeSCahill:
I think this needs the Patron font set: https://github.com/redpanda-data/documentation/blob/a6e07cb5692b5477329fb4590cfc7ff3f2e5b501/src/css/custom.css#L22

Issue #1170, opened 2023-02-07
Add rolling restart guide for Kubernetes deployments to docs


vuldin:
### Describe the Issue

I Redpanda user requested details on how to do rolling restarts for a Redpanda cluster within Kubernetes, and I didn't see any documentation on this. We have the following two docs:
https://docs.redpanda.com/docs/manage/cluster-maintenance/rolling-upgrade/
https://docs.redpanda.com/docs/manage/node-management/

Each covers a portion of the process for rolling restarts, but neither gives details specific to Kubernetes deployments.

### Updates to existing documentation

It's possible either of the links above could be modified to include the details I've included in this issue.

### New feature or content gap requests

It's also possible this is a content gap that deserves it's own new page. We have a Redpanda University course that covers this content fairly well, but again it does not mention Kubernetes.

### Additional notes

Below is what I have prepared for the Redpanda user who requested this info, and the documentation change could incorporate this content:

===

### Externally connect via local rpk
In order to use `rpk` from your local computer, you need to know how to access the Kubernetes cluster externally. Find your external advertised broker addresses and ports by first running the following command:
```
kubectl exec -it -n redpanda redpanda-0 -c redpanda -- rpk cluster info --brokers localhost:9092
```

Output:
```
CLUSTER
=======
redpanda.c2b00813-0c97-4622-a377-236f593ec915

BROKERS
=======
ID    HOST        PORT
0*    redpanda-0  31092
1     redpanda-1  31092
2     redpanda-2  31092
```

In the above output we see that each broker advertises port 31092 (which is the default port when using the NodePort service).

> Note: Replace `9092` with whatever you have defined for the external kafka API traffic. If you have deployed Redpanda via the helm chart, it defaults to `9094`.

> Note: Make sure that you aren't mistakenly sending the above command to the port for *internal* kafka API traffic! If you have deployed Redpanda via the helm chart, it defaults to `9093`.

It is helpful to set `REDPANDA_BROKERS` for rpk commands that use the kafka API:
``` 
export REDPANDA_BROKERS=redpanda-0:31092,redpanda-1:31092,redpanda-2:31092
```

You should now be able to use your local `rpk` to access your cluster. If you run into issues, see this [link](https://docs.redpanda.com/docs/manage/kubernetes/networking/networking-and-connectivity/) for more details and help.

### Following logs
[...]

Issue #1061, opened 2023-01-19
Describe which metrics to watch for a successful rolling upgrade


JakeSCahill:
### Describe the Issue

From @RafalKorepta in Slack:

I just happen to bump into [rolling upgrade docs](https://docs.redpanda.com/docs/manage/cluster-maintenance/rolling-upgrade/). Could we add, between points 2 and 3, a description of metrics to watch to make upgrades more robust?

- no longer leaderless partitions -> vectorized_cluster_partition_leader_id == -1
- IO metrics are similar prior to the beginning of the rolling restart
- vectorized_cluster_partition_under_replicated_replicas == 0
- 
You can read the PR cover letter https://github.com/redpanda-data/redpanda/pull/7594 and this [comment](https://github.com/redpanda-data/redpanda/issues/3023#issuecomment-1323371120).

Link to the original conversation: https://redpandadata.slack.com/archives/C01H6JRQX1S/p1674121816591449?thread_ts=1674064812.393709&cid=C01H6JRQX1S

### Updates to existing documentation

[Rolling Upgrades](https://docs.redpanda.com/docs/manage/cluster-maintenance/rolling-upgrade/)

### New feature or content gap requests

Content gap.

Feediver1:
Not sure who should take this on - @shorgi or @JakeSCahill ? I suspect @shorgi  can provide the metrics to monitor to @JakeSCahill for this? Please let me know if any issues. Thx.

Issue #415, opened 2022-05-24
Add some important caveats in the Upgrade Guide


daisukebe:
### Who is this for and what problem do they have today?

Current `Version upgrade` guide doesn't mention about the following important caveats, specifically for upgrade from 21.11 to 22.1:

* Group topic migration
* Downgrade capability

### What would you like fixed in or added to the documentation?

We should be very clear that:

* As soon as all nodes are up and running with 22.1, it starts to migrate the consumer group topics to `__consumer_offsets`, which is a new feature introduced in 22.1 for Kafka compatibility. It will take some time depending on the data size, and until it finishes all consumer group related operations (consume, offset commit, coordinator election, etc) will be blocked.

* Once the cluster is upgraded to 22.1, there's no supported way to downgrade to the previous major version, 21.x.y. 

### Is this a new feature or a gap in the documentation?

Gap

### Link to existing product documentation page and/or community slack conversation, if applicable

https://docs.redpanda.com/docs/cluster-administration/version-upgrade/

### Additional notes

We've learned this from the latest incident with Jump Trading, at #inc-2022-05-21-jumptrading-data-read-stopped-after-221. Will post more details soon.

Feediver1:
This should go into https://github.com/redpanda-data/documentation/issues/1010

Closing this as a duplicate, but adding the request to https://github.com/redpanda-data/documentation/issues/1010.

PR #1246, opened 2023-02-16
docs:#1010 combine version upgrades into rolling upgrades

### LINKED ISSUE #1010: Combine versions upgrade and rolling upgrade docs
### Describe the Issue

Address comment in #960 (https://github.com/redpanda-data/documentation/pull/960#pullrequestreview-1242231266):

- Because the [versions upgrade](https://docs.redpanda.com/docs/platform/install-upgrade/version-upgrade/) and [rolling upgrade](https://docs.redpanda.com/docs/platform/install-upgrade/rolling-upgrade/) docs are interrelated, investigate combining them into a single page to improve their readability and usability.

Combine [versions upgrade](https://docs.redpanda.com/docs/platform/install-upgrade/version-upgrade/) and [rolling upgrade](https://docs.redpanda.com/docs/platform/install-upgrade/rolling-upgrade/) docs.

micheleRP:
resolves #1010 

This incorporates the Version Upgrades content into the Rolling Upgrades page. (Version Upgrades was removed from our new library structure, so this adds the upgrade info back in.) It also adds some other upgrade info/tips that was missing from our doc. 
- preview 23.1: https://deploy-preview-1246--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/rolling-upgrade/
- preview 22.3: https://deploy-preview-1246--redpanda-documentation.netlify.app/docs/22.3/manage/cluster-maintenance/rolling-upgrade/


micheleRP:
@vuldin, @daisukebe, @dotnwat: Please review this revised page for rolling upgrades. Also, I know we've removed many instances of "vectorized", so please confirm also if it's OK to use:

If you previously installed Redpanda with brew, run:
`brew install vectorizedio/tap/redpanda`

and 
```yaml
...
spec:
  image: "vectorized/redpanda"
  version: "<current_version>"
...
```

JakeSCahill:
@micheleRP I think we can take Kubernetes info out of this page. It’s confusing to see maintenance mode tasks when they don’t apply. I’ll be doing a dedicated Kubernetes page as part of https://github.com/redpanda-data/documentation/issues/1170 this sprint.

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -7,33 +7,117 @@ title: Rolling Upgrades
     <meta name="description" content="One of the primary uses for maintenance mode is to perform a rolling upgrade on each node in the cluster."/>
 </head>
 
-One of the primary uses for maintenance mode is to perform a rolling upgrade on each node in the cluster. This process involves putting a node into maintenance mode, upgrading the node, taking the node out of maintenance mode, and then repeating the process on the next node in the cluster. Placing nodes into maintenance mode ensures a smooth upgrade of your cluster while reducing the risk of interruption or degradation in service.
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+To benefit from Redpanda's new features and enhancements, upgrade to the latest version. New features are enabled after all nodes (brokers) in the cluster are upgraded.
```
daisukebe, line 13: Can we add this? New features are enabled as long as `all` nodes are alive with a given new version. In a case there's a dead node in the cluster, enabling new features won't be triggered, which is not explicit to users.

(forgive me my broken English :wink: )
```
Ensure all nodes are up and running before upgrading the version. Use 'rpk redpanda admin brokers list'.
```

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -7,33 +7,117 @@ title: Rolling Upgrades
...
+
+:::note notes
```
daisukebe, line 15: Do we have a definition on; what `one version` mean and what `minor versions` mean by?

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -7,33 +7,117 @@ title: Rolling Upgrades
...
+
+## Find your current version
+
+To determine your current version, run:
+
+<Tabs>
+<TabItem value="check-linux" label="Linux" default>
+
+```bash
+curl -s -XGET http://<broker host>:9644/v1/brokers/
```
daisukebe, line 31: Assuming all users are running >= v22.1, no more v21.x users, users can identify the broker version using `rpk redpanda admin brokers list`. @mattschumpert, can we assume so?

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -7,33 +7,117 @@ title: Rolling Upgrades
...
+```bash
+brew list --versions | grep redpanda
+```
+
+</TabItem>
+
+<TabItem value="check-kubernetes" label="Kubernetes" default>
+
+```bash
+kubectl exec -it -n <cluster_name> <pod-name> -- rpk version
```
daisukebe, line 51: Same above. Assuming all users are running >= 22.1, we can use the command below.

```
kubectl exec -it -n <cluster_name> <pod-name> -- rpk redpanda admin brokers list
```

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: Great question, @daisukebe. @mattschumpert, and if we have no v21.11 users, can we remove all 21.11 doc; e.g., https://docs.redpanda.com/docs/21.11/home/

micheleRP:
> #1170 this sprint.

Thanks, @JakeSCahill: I've removed K8s content from this page. Please remind me after your content is in to add a link to it on this page (if you don't).

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: Not sure about customers running 21.11. See https://redpandadata.slack.com/archives/C02JDTMG8GN/p1677096032290669

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: Thanks @daisukebe! Please review this updated intro: https://deploy-preview-1246--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/rolling-upgrade/

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: Thanks, @daisukebe! I added this in the Perform a rolling upgrade section intro. Please review: https://deploy-preview-1246--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/rolling-upgrade/#perform-a-rolling-upgrade

micheleRP:
> @vuldin, @daisukebe, @dotnwat: Please review this revised page for rolling upgrades. Also, I know we've removed many instances of "vectorized", so please confirm also if it's OK to use:
> 
> If you previously installed Redpanda with brew, run: `brew install vectorizedio/tap/redpanda`
> 
> and
> 
> ```yaml
> ...
> spec:
>   image: "vectorized/redpanda"
>   version: "<current_version>"
> ...
> ```

@JakeSCahill The 2nd reference here is moving to your new K8s page. For the 1st reference: I changed this to `brew install redpanda-data/tap/redpanda`. Can you please confirm that that's correct?

JakeSCahill:
> > @vuldin, @daisukebe, @dotnwat: Please review this revised page for rolling upgrades. Also, I know we've removed many instances of "vectorized", so please confirm also if it's OK to use:
> > If you previously installed Redpanda with brew, run: `brew install vectorizedio/tap/redpanda`
> > and
> > ```yaml
> > ...
> > spec:
> >   image: "vectorized/redpanda"
> >   version: "<current_version>"
> > ...
> > ```
> 
> @JakeSCahill The 2nd reference here is moving to your new K8s page. For the 1st reference: I changed this to `brew install redpanda-data/tap/redpanda`. Can you please confirm that that's correct?

`brew install redpanda-data/tap/redpanda` works

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -68,7 +165,94 @@ Repeat this process for each node to upgrade.
...
+docker restart <container_name>
+```
+
+For more information, see [Redpanda Quickstart](../../../get-started/quick-start/quick-start-docker).
+
+</TabItem>
+
+</Tabs>
+
+### Disable maintenance mode
```
micheleRP, line 255: @RafalKorepta Issue #1061 says to add a description here of what to metrics to monitor at this step. Our doc hasn't yet documented vectorized_cluster_partition_leader_id or vectorized_cluster_partition_under_replicated_replicas, and I'm not sure which I/O metrics to add here. What general recommendations should we include (and then we can add links to specific metrics/APIs when we have doc for them)?

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
daisukebe: Great, thanks!

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
mattschumpert: We should have paying customers off 21.11 docs now, but we should only mark them as EOL at the top of the pages not remove them entirely because its unfriendly to community users on that version who still want to use community edition and/or upgrade eventually.

mattschumpert: However, for current version doc set, we can make the assumption of rpk redpanda admin brokers list for instructional purposes.

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -7,33 +7,112 @@ title: Rolling Upgrades
...
+</TabItem>
+
+<TabItem value="check-docker" label="Docker" default>
+
+:::caution
+Running Redpanda directly on Docker is not supported for production usage. This platform should only be used for testing.
+:::
+
+```bash
+docker exec -it <container_name> rpk version
```
RafalKorepta, line 57: I would be nit pick this command. The container_name variable should consist of 

1. container name
2. container tag
The container tag would determine what version of rpk is used. Not necessary the Redpanda binary. Fortunately the release process is construct in that way we bundle rpk and Redpanda into the same container tag with the same version. 

If you would like to list the available docker hub tags the following command:
```bash
curl -s 'https://hub.docker.com/v2/repositories/vectorized/redpanda/tags/?ordering=last_updated&page=1&page_size=50' | jq -r '.results[].name'
```  
would return:
```
v22.3.13
latest
v22.3.13-arm64
v22.3.13-amd64
v22.2.10
v22.2.10-arm64
v22.2.10-amd64
v22.3.12
v22.3.12-arm64
v22.3.12-amd64
v22.3.11
v22.3.11-arm64
v22.3.11-amd64
v22.3.10
v22.3.10-arm64
v22.3.10-amd64
v22.1.11
v22.1.11-arm64
v22.1.11-amd64
v22.2.9
v22.2.9-arm64
v22.2.9-amd64
v22.3.9
v22.3.9-arm64
v22.3.9-amd64
v22.3.8
v22.3.8-arm64
v22.3.8-amd64
v22.3.6
v22.3.6-arm64
v22.3.6-amd64
v22.3.5
v22.3.5-arm64
v22.3.5-amd64
v22.2.8
v22.2.8-arm64
v22.2.8-amd64
v22.3.4
v22.3.4-arm64
v22.3.4-amd64
v22.1.10
v22.1.10-arm64
v22.1.10-amd64
v21.11.22
latest-arm64
v21.11.22-arm64
latest-amd64
v21.11.22-amd64
v21.11.22-beta2
v21.11.22-beta2-arm64
```

You can always adjust the number of result using `page_size`.

If our user would like to see v22 available version, they could use the following command:
```bash
curl -s 'https://hub.docker.com/v2/repositories/vectorized/redpanda/tags/?name=v22&ordering=last_updated&page=1&page_size=25' | jq -r '.results[].name'
```

The name parameter `v22` could be changed to what ever release train we have e.g. `v23` or `v22.3.`

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -7,33 +7,112 @@ title: Rolling Upgrades
...
-### 1. Enable maintenance mode
+
+## Perform a Rolling Upgrade
+
+Placing nodes into [maintenance mode](../../node-management) ensures a smooth upgrade of your cluster while reducing the risk of interruption or degradation in service. A rolling upgrade involves putting a node into maintenance mode, upgrading the node, taking the node out of maintenance mode, and then repeating the process on the next node in the cluster. 
+
+To ensure that all nodes are active before upgrading, run:
+
+```bash 
+rpk redpanda admin brokers list
```
RafalKorepta, line 89: Maybe we should use 
```
rpk cluster health
```

This command would return:
```
CLUSTER HEALTH OVERVIEW
=======================
Healthy:               true
Controller ID:               0
All nodes:                   [1 2 0]
Nodes down:                  []
Leaderless partitions:       []
Under-replicated partitions: []
```

This command would be available starting from v22.1.1

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -7,33 +7,112 @@ title: Rolling Upgrades
...
+```
+
+The logical cluster version is incremented after the last active node is upgraded. If problems occur, the upgrade is not committed.
+
+:::note
+After all nodes are upgraded, Redpanda starts to migrate consumer group topics to __consumer_offsets. This takes some time, depending on the data size. Until it finishes, all consumer group related operations (consume, offset commit, coordinator election) are blocked.
```
RafalKorepta, line 95: This apply only to upgrades from v21.11 to v22.1 according to the @daisukebe issue. REF https://github.com/redpanda-data/documentation/issues/415

> As soon as all nodes are up and running with 22.1, it starts to migrate the consumer group topics to __consumer_offsets, which is a new feature introduced in 22.1 for Kafka compatibility. It will take some time depending on the data size, and until it finishes all consumer group related operations (consume, offset commit, coordinator election, etc) will be blocked.


I'm not sure if this should be precisely mention `from v21.11.x to v22.1.x`

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
RafalKorepta: Oh I see that it's addressed in the next headline `Enable maintenance mode`

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -7,33 +7,112 @@ title: Rolling Upgrades
...
     ```
 
-    If the cluster is healthy, the output shows `Healthy: true`.
+    <details>
+    <summary>
+    Expected output
+    </summary>
+
+    ```bash
+    Healthy: true
```
RafalKorepta, line 112: NIT:

Could we show whole output but highlight only the `Healthy: true`?

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -48,16 +127,34 @@ Repeat this process for each node to upgrade.
...
-    
-    Confirm that `FINISHED` is `true`.
+
+    <details>
+    <summary>
+    Expected output
+    </summary>
+
+    ```bash
+    FINISHED: true
```
RafalKorepta, line 137: ditto

whole output with highlight of `FINISHED: true`

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -48,16 +127,34 @@ Repeat this process for each node to upgrade.
...
     ```
 
-    If the cluster is healthy, the output shows `Healthy: true`.
+    <details>
+    <summary>
+    Expected output
+    </summary>
+
+    ```bash
+    Healthy: true
```
RafalKorepta, line 154: ditto

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -68,7 +165,94 @@ Repeat this process for each node to upgrade.
...
+</Tabs>
+
+
+</TabItem>
+<TabItem value="upgrade-macos" label="MacOS" default>
+
+If you previously installed Redpanda with brew, run:
+
+```bash
+brew install redpanda-data/tap/redpanda
```
RafalKorepta, line 202: I'm not sure if `install` is the right command for an upgrade. I'm personally used to 
```
brew upgrade ...
```

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -68,7 +165,94 @@ Repeat this process for each node to upgrade.
...
+docker pull docker.redpanda.com/redpandadata/redpanda:<version>
+```
+
+After it completes, restart the cluster:
+
+```bash
+docker restart <container_name>
+```
+
+For more information, see [Redpanda Quickstart](../../../get-started/quick-start/quick-start-docker).
```
RafalKorepta, line 249: I'm not sure if we should advise user to use docker compose or even `rpk container start` for testing:
```
rpk container start --help
Start a local container cluster

Usage:
  rpk container start [flags]

Flags:
  -h, --help           Help for start
  -n, --nodes uint     The number of nodes to start (default 1)
      --retries uint   The amount of times to check for the cluster before considering it unstable and exiting. (default 10)

Global Flags:
  -v, --verbose   Enable verbose logging (default: false)
```

or 
https://github.com/redpanda-data/documentation/blob/e289b8f91e617be77c2313ccc734c2742f4a7966/docs/get-started/quick-start.mdx?plain=1#L121-L191
or
https://github.com/redpanda-data/documentation/blob/e289b8f91e617be77c2313ccc734c2742f4a7966/docs/get-started/quick-start.mdx?plain=1#L225-L355

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -84,10 +268,26 @@ After you've successfully upgraded the node:
     rpk cluster maintenance status
     ```
 
-### 3. Restart node
+### Restart node
```
RafalKorepta, line 271: I'm a bit confused. The restart should be before disabling maintenance mode.

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
RafalKorepta: Thanks to @jcsp ([PR-9011](https://github.com/redpanda-data/redpanda/pull/9011)) the cluster health will have more useful information starting from v23.1.x.

Before this version I would suggest to watch for:

- any none zero value for metrics `vectorized_cluster_partition_under_replicated_replicas`. That would indicate replication can not catch up for various reasons and the upgrade should be paused.
- if vectorized_cluster_partition_leader_id for any partition has -1 value. That would indicate we lost partition leader and we can not produce to such partition
- IO metrics:
  - vectorized_storage_log_written_bytes - The rate for each broker should go to circa value before restart
  - vectorized_storage_log_read_bytes - The rate for each broker should go to circa value before restart
  - vectorized_raft_leadership_changes - after restart eventually it should go to 0
  - vectorized_raft_leader_for - the sum for each broker should be equal. Eventually
  - vectorized_kafka_latency_produce_latency_us_bucket - p99 histogram should go to circa value before restart
  - vectorized_kafka_latency_fetch_latency_us_bucket - p99 histogram should go to circa value before restart
  - vectorized_kafka_rpc_dispatch_handler_latency_bucket - p99 histogram should go to circa value before restart


If @jcsp or @esteban have more details to share please feel free to correct me :)

RafalKorepta (commented):
I left few comments

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
daisukebe: > I'm not sure if this should be precisely mention from v21.11.x to v22.1.x

Agreed.

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: resolved here: https://github.com/redpanda-data/documentation/pull/1297/files#diff-2fac7b96f96c7e0bb400f0a2ee80bc5bce9b3841860bc637189110ef6ed406d4

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: Thanks @RafalKorepta and @daisukebe. Is this note still relevant to current customers (e.g., customers upgrading to 23.1), or should we remove the note altogether? 

After all nodes are upgraded, Redpanda starts to migrate consumer group topics to __consumer_offsets. This takes some time, depending on the data size. Until it finishes, all consumer group related operations (consume, offset commit, coordinator election) are blocked.

daisukebe: @micheleRP , we may want to leave it as is to be friendly for users still running 21.11 or lower. However, the logic only applies to the one time upgrade from 21.11 to 22.1. We should note it. When upgrading from 22.1 to 22.2 and further, this isn't the case any more.

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: Thanks @RafalKorepta! Please check this new step: https://deploy-preview-1246--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/rolling-upgrade/#check-metrics. And a couple questions:
- I don't understand "go to circa value". Could I change this "return to the original value"? If so, should users check these metrics before running the upgrade too? 
- This is going out with the 23.1 docs. Can you please confirm that we should include this section (i.e., that checking `rpk cluster health` is not enough)?

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: Thanks @RafalKorepta! Please review https://deploy-preview-1246--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/rolling-upgrade/#find-your-current-version

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: Thanks @RafalKorepta. Please confirm if this now looks correct: https://deploy-preview-1246--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/rolling-upgrade/#perform-a-rolling-upgrade

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: @RafalKorepta I added more expected output many of the steps. Let me know if it's too much!

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: for reference: https://redpandadata.slack.com/archives/C04LYNJQUGN/p1677794110689209

JakeSCahill (approved):
Other than the existing open questions in the comments, LGTM. @micheleRP we should add the impact of restarts partial to this doc: https://github.com/redpanda-data/documentation/tree/dev/docs/manage/shared/rolling-upgrades

micheleRP:
> Other than the existing open questions in the comments, LGTM.
> 
> @micheleRP we should add the impact of restarts partial to this doc: https://github.com/redpanda-data/documentation/tree/dev/docs/manage/shared/rolling-upgrades
> 
> Also, do we need to forwardport these changes to 23.1?
> 
> The build is failing ATM because:
> 
> ```
> 7:30:50 PM: - On source page path = /docs/22.3/manage/cluster-maintenance/rolling-upgrade/:
> 7:30:50 PM:    -> linking to ../../../get-started/quick-start/quick-start-docker/ (resolved as: /docs/22.3/get-started/quick-start/quick-start-docker/)
> ```

Thanks, @JakeSCahill! I added the shared file to 22.3 and also made some more edits to be consistent with your K8s version.

micheleRP:
@mattschumpert, @RafalKorepta, @daisukebe, @vuldin, @dotnwat: This PR now also incorporates the updates in https://github.com/redpanda-data/documentation/pull/1297. Please give this a final review!

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
RafalKorepta: > Could I change this "return to the original value"? If so, should users check these metrics before running the upgrade too?

The `original value` for me doesn't bring the context that value should come back to the value before an upgrade. Never the less my English is not that good :) 

> This is going out with the 23.1 docs. Can you please confirm that we should include this section (i.e., that checking rpk cluster health is not enough)?

The `rpk cluster health` is not enough as I think metrics should be watched too. I will let @jcsp confirm this.

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
RafalKorepta: Great!

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
RafalKorepta: LGTM

versioned_docs/version-22.3/manage/shared/rolling-upgrades/_restart-impact.mdx
```diff
@@ -0,0 +1,18 @@
...
+When you restart one or more brokers in a cluster, clients (consumers and producers) may experience higher latency due to partition leadership reassignment. Because clients must communicate with the leader of a partition, they may send a request to a broker whose leadership has been transferred, and receive `NOT_LEADER_FOR_PARTITION`. In this case, clients must request metadata from the cluster to find out the address of the new leader. Clients refresh their metadata periodically, or when the client receives some retryable errors that indicate that the metadata may be stale. For example:
+
+1. Broker A shuts down.
+1. Client sends a request to broker A, and receives `NOT_LEADER_FOR_PARTITION`.
+1. Client requests metadata, and learns that the new leader is broker B.
+1. Client sends the request to broker B.
+
+### CPU spikes upon broker restart
+
+When a restarted broker becomes available again, you may see your nodes' CPU usage increase temporarily. This temporary increase in CPU usage is due to the cluster rebalancing the partition replicas.
```
RafalKorepta, line 14: Recent incident brought good point that production setup for Redpanda will always use around 80-95% CPU as [Seastar](https://seastar.io/) is always pooling. It might be misleading for users. They should look for `vectorized_reactor_utilization` metrics. @daisukebe could know if there is more dimensions/metrics to watch for.

jcsp:
Everywhere this page mentions a `vectorized_` metric, it should be mentioning a `redpanda_` metric (or omit it if there is no public metric) -- we do not want to encourage users to use the internal metrics endpoint.

micheleRP:
> > If @jcsp or @esteban have more details to share please feel free to correct me :)
> 
> Everywhere this page mentions a `vectorized_` metric, it should be mentioning a `redpanda_` metric (or omit it if there is no public metric) -- we do not want to encourage users to use the internal metrics endpoint.

Thanks @jcsp. Would you please confirm if this looks correct: https://deploy-preview-1246--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/rolling-upgrade/#check-metrics

versioned_docs/version-22.3/manage/shared/rolling-upgrades/_restart-impact.mdx
micheleRP: and please see John's comment below that we shouldn't encourage users to use the internal metrics endpoint. Should they check [redpanda_uptime_seconds_total ](https://docs.redpanda.com/docs/reference/public-metrics-reference/#redpanda_uptime_seconds_total) instead?

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,425 @@ contextLinks:
...
+
+## Find a new version
+
+Before you perform a rolling upgrade, you must find out which Redpanda version you are currently running, whether you can upgrade straight to the new version, and what's changed since your original version.
+
+Redpanda version numbers follow the convention AB.C.D, where AB is the two digit year, C is the feature release, and D is the patch release. For example, version 22.3.1 indicates the first patch release on the third feature release of the year 2022. 
+
+:::note notes
+- Redpanda only supports upgrading one sequential feature release at a time. For example, you can upgrade from the 22.2 feature release to 22.3. You cannot skip feature releases. 
+- Redpanda only supports downgrading between sequential patch releases. For example, you can downgrade from the 22.2.2 patch release to 22.2.1, but you cannot downgrade to 22.1.7.
```
hcoyote, line 49: Thes two statements imply that if we upgrade from 22.1.<latest> to 22.2.latest to move between feature releases, that we cannot do a rollback in the event we need to do a downgrade. Meaning, we can only roll-forward. That doesn't seem right and will present issues to customers.

versioned_docs/version-22.3/manage/shared/rolling-upgrades/_restart-impact.mdx
```diff
@@ -0,0 +1,18 @@
+When brokers restart, clients may experience higher latency, nodes may experience CPU spikes when the broker becomes available again, and you may receive alerts about under-replicated partitions.
```
hcoyote, line 1: Related to comment earlier about single-replica topics. The impact of restarts should also call out that topics using replication factor = 1 would be unavailable. I would classify this as somewhat different than under-replicated partitions. 

we make a comment in rolling-upgrade.mdx about repl=1 and probably need to address that here as well. While it's not something we recommend people configuring for, if they do it, they'll end up with a data access outage.

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -4,90 +4,421 @@ title: Rolling Upgrades
...
+
+:::caution
+Running Redpanda directly on Docker is not supported for production usage. This platform should only be used for testing.
+:::
+
+```bash
+docker exec -it <container_name><container_tag> rpk version 
+```
+
+Remember to replace the variables `<container_name>` and `container_tag`. The container tag determines which version of `rpk` to use. The release process bundles `rpk` and `Redpanda` into the same container tag with the same version.
```
hcoyote, line 71: this should probably be `<container_tag>` for formatting consistency with the command line above.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,425 @@ contextLinks:
...
+
+:::caution
+Running Redpanda directly on Docker is not supported for production usage. This platform should only be used for testing.
+:::
+
+```bash
+docker exec -it <container_name><container_tag> rpk version 
+```
+
+Remember to replace the variables `<container_name>` and `container_tag`. The container tag determines which version of `rpk` to use. The release process bundles `rpk` and `Redpanda` into the same container tag with the same version.
```
hcoyote, line 82: this should probably be <container_tag> for formatting consistency with the command line above.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,425 @@ contextLinks:
...
+v22.2.10-arm64
+v22.2.10-amd64
+v22.3.12
+v22.3.11
+v22.3.10
+...
+```
+
+</details>
+
```
hcoyote, line 128: This only makes sense in the context of a docker setup. If you're looking at the Linux or MacOS tabs, the available version format strings are going to be different because it's going to depend on the OS package management tool.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,425 @@ contextLinks:
...
-## Performing a rolling upgrade
+## Perform a rolling upgrade
 
-To perform a rolling upgrade on a node in the cluster:
+Placing brokers into [maintenance mode](../../node-management) ensures a smooth upgrade of your cluster while reducing the risk of interruption or degradation in service. A rolling upgrade involves putting a broker into maintenance mode, upgrading the broker, taking the broker out of maintenance mode, and then repeating the process on the next broker in the cluster. 
 
-1. [Enable maintenance mode](#1-enable-maintenance-mode).
-2. [Disable maintenance mode](#2-disable-maintenance-mode).
-3. [Restart the node](#3-restart-node).
+To ensure that all brokers are active before upgrading, run:
```
hcoyote, line 137: probably want to note that they need to look at the Membership-Status column, just in case none of those entries say "active". A pointer to what to do if they're not active might be good as well.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,425 @@ contextLinks:
...
+```
+NODE-ID  NUM-CORES  MEMBERSHIP-STATUS  IS-ALIVE  BROKER-VERSION
+0        1          active             true      v22.3.11 - 9eefb907c43bf1cfeb0783808c224385c857c0d4-dirty
+1        1          active             true      v22.3.11 - 9eefb907c43bf1cfeb0783808c224385c857c0d4-dirty
+2        1          active             true      v22.3.11 - 9eefb907c43bf1cfeb0783808c224385c857c0d4-dirty
+```
+
+</details>
+
+The logical cluster version is incremented after the last active broker is upgraded. If problems occur, the upgrade is not committed.
```
hcoyote, line 157: Is this talking about the broker-version in the previous output or the idea of the cluster config version?

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,425 @@ contextLinks:
...
+
+:::note
+When upgrading from version 21.11 to 22.1, after all brokers are upgraded, Redpanda starts to migrate consumer group topics to __consumer_offsets. This takes some time, depending on the data size. Until it finishes, all consumer group related operations (consume, offset commit, coordinator election) are blocked.
```
hcoyote, line 160: How can they track this or know when it's completed?

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,425 @@ contextLinks:
...
+| [redpanda_cluster_unavailable_partitions](../../../reference/public-metrics-reference/#redpanda_cluster_unavailable_partitions)  | If any partition shows a value of `-1`, then the partition leader is lost, and Redpanda cannot produce it.   |  
+| [redpanda_kafka_request_bytes_total](../../../reference/public-metrics-reference/#redpanda_kafka_request_bytes_total)| Before restart, the product and consume rate for each broker should go to circa value. |
+| [redpanda_raft_leadership_changes](../../../reference/public-metrics-reference/#redpanda_raft_leadership_changes)  | After restart, this value should eventually go to 0. |      
+| [redpanda_kafka_request_latency_seconds](../../../reference/public-metrics-reference/#redpanda_kafka_request_latency_seconds) | Before restart, the p99 histogram should go to circa value.    |   
+| [redpanda_rpc_request_latency_seconds](../../../reference/public-metrics-reference/#redpanda_rpc_request_latency_seconds) | Before restart, the p99 histogram should go to circa value.    |  
+
+### Restart broker
+
+After you're out of maintenance mode, restart the broker's Redpanda service.
```
hcoyote, line 373: Wouldn't we restart this before we exit maintenance mode?

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,425 @@ contextLinks:
...
+
+:::note
+You can stop the upgrade process and roll back to the original version as long as you have not upgraded every broker and restarted the cluster.
```
hcoyote, line 376: This seems like an important caveat that we should list higher up in the doc. Maybe in the initial intro section or under the pre-reqs as part of the understanding people need to work through.

hcoyote (commented):
I ran through the generic rolling-upgrade.mdx and skimmed over the 22.3 version ... probably everything I commented on in the generic applies to the versioned doc as well.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: @daisukebe for this use case, do you know how to track when consumer groups are done migrating?

docs/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: @mattschumpert can you please confirm if this note is correct. And if so, if Travis's comment is also correct?

versioned_docs/version-22.3/manage/shared/rolling-upgrades/_restart-impact.mdx
micheleRP: @hcoyote do you think we should just add to the end of that paragraph: "Topics that weren't using replication (that is, topics that had `replication.factor=1`) will be unavailable." @JakeSCahill since you worked on this!

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,425 @@ contextLinks:
...
-One of the primary uses for maintenance mode is to perform a rolling upgrade on each node in the cluster. This process involves putting a node into maintenance mode, upgrading the node, taking the node out of maintenance mode, and then repeating the process on the next node in the cluster. Placing nodes into maintenance mode ensures a smooth upgrade of your cluster while reducing the risk of interruption or degradation in service.
+To benefit from Redpanda's new features and enhancements, upgrade to the latest version. New features are enabled after all brokers (nodes) in the cluster are upgraded. 
+
+Rolling upgrades require all brokers to be placed into [maintenance mode](../../node-management) and restarted separately, one after the other. Placing brokers into maintenance mode ensures a smooth upgrade of your cluster while reducing the risk of interruption or degradation in service. When a broker is placed into maintenance mode, it reassigns its partition leadership to other brokers for all topics that have a replication factor greater than one. Reassigning partition leadership involves _draining_ leadership from the broker and _transferring_ that leadership to another broker.
```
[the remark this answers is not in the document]
micheleRP: @hcoyote I believe all the caveats we know about are now in this doc! Should we recommend that for topics with `replication.factor=1`, they increase their replication factor before upgrade? cc @JakeSCahill

docs/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: @hcoyote what should they do if one is not active?

docs/manage/cluster-maintenance/rolling-upgrade.mdx
daisukebe: Good question. Once you see `consumer offset feature enabled` in all nodes, we can assume it's done.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,427 @@ contextLinks:
...
 
-2. Select a node that has not been upgraded yet and place it into maintenance mode: 
+2. Select a broker that has not been upgraded yet and place it into maintenance mode: 
 
     ```bash
     rpk cluster maintenance enable <node-id> --wait
     ```
 
-    The `--wait` option ensures that the cluster is healthy before putting the node into maintenance mode.
+    The `--wait` option ensures that the cluster is healthy before putting the broker into maintenance mode. This example puts broker 0 into maintenance mode.
```
daisukebe, line 201: > The `--wait` option ensures that the cluster is healthy before putting the broker into maintenance mode.

This is not accurate. Rather we should say something like below.

> The `--wait` option leaves the command as is and waits until a given node, 0 in this example, finishes draining all partitions it originally served. Once the partition draining completes, the command finished.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
daisukebe: It is talking about a Redpanda internal logical version, which tracks new features. It's not either a broker-version or a cluster config version at all. I guess it's not a good idea to mention about `logical cluster version` unless we explicitly explains it somewhere. It just confuses users. Rather can we just say something like below? @jcsp 

```
A new product feature, that's available in a given feature release, will be fully activated after the last active broker is upgraded. If problems occur, the upgrade is not committed.
```

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,427 @@ contextLinks:
...
-3. [Restart the node](#3-restart-node).
+To ensure that all brokers are active before upgrading, run:
 
-Repeat this process for each node to upgrade.
+```bash 
+rpk redpanda admin brokers list
+```
 
-### 1. Enable maintenance mode
+All brokers should list `active` in the `MEMBERSHIP-STATUS` column:
```
daisukebe, line 147: > All brokers should list `active` in the `MEMBERSHIP-STATUS` column:

and should list `true` in the `IS-ALIVE` column too

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,427 @@ contextLinks:
...
+</Tabs>
+
+
+<details>
+<summary>
+Example output
+</summary>
+
+```bash
+v22.3.11 (rev 9eefb90)
```
daisukebe, line 99: The example output of `rpk redpanda admin brokers list` in Linux is below, which differs from the commands in MacOS and Docker. 

```
NODE-ID  NUM-CORES  MEMBERSHIP-STATUS  IS-ALIVE  BROKER-VERSION
0        1          active             true      v22.3.11 - 9eefb907c43bf1cfeb0783808c224385c857c0d4-dirty
1        1          active             true      v22.3.11 - 9eefb907c43bf1cfeb0783808c224385c857c0d4-dirty
2        1          active             true      v22.3.11 - 9eefb907c43bf1cfeb0783808c224385c857c0d4-dirty
```

docs/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: I updated this from: 
The logical cluster version is incremented after the last active broker is upgraded.
to 
New features in a version are enabled after all brokers in the cluster are upgraded.

versioned_docs/version-22.3/manage/shared/rolling-upgrades/_restart-impact.mdx
daisukebe: @jcsp , should we encourage users to monitor `redpanda_cpu_busy_seconds_total` instead of `vectorized_reactor_utilization`? If that's the case, do we recommend them to take the derivative as you mentioned in https://redpandadata.slack.com/archives/C01ND4SVB6Z/p1669395833531899?thread_ts=1669392271.380349&cid=C01ND4SVB6Z?

docs/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -11,95 +11,427 @@ contextLinks:
...
+### Check metrics
+
+Check the following metrics before continuing with the upgrade:
+
+| Metric | Description | 
+| ---               | ---          
+| [redpanda_kafka_under_replicated_replicas](../../../reference/public-metrics-reference/#redpanda_kafka_under_replicated_replicas) | If this shows any non-zero value, then replication cannot catch up, and the upgrade should be paused. |
+| [redpanda_cluster_unavailable_partitions](../../../reference/public-metrics-reference/#redpanda_cluster_unavailable_partitions)  | If any partition shows a value of `-1`, then the partition leader is lost, and Redpanda cannot produce to it.   |  
+| [redpanda_kafka_request_bytes_total](../../../reference/public-metrics-reference/#redpanda_kafka_request_bytes_total)| Before restart, the produce and consume rate for each broker should go to circa value. |
+| [redpanda_raft_leadership_changes](../../../reference/public-metrics-reference/#redpanda_raft_leadership_changes)  | After restart, this value should eventually go to 0. |
```
jcsp, line 373: This is a counter: it won't go to zero, but it will stop changing.

I don't know that we necessarily need to ask people to do this though.  It is okay to proceed with an upgrade while some leader transfers are happening.

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
```diff
@@ -4,90 +4,423 @@ title: Rolling Upgrades
...
+<TabItem value="check-linux" label="Linux" default>
+
+```bash
+rpk redpanda admin brokers list
+```
+
+For all available flags, see the [`rpk redpanda admin brokers list` command reference](../../../reference/rpk/rpk-redpanda/rpk-redpanda-admin-brokers-list).
+
+</TabItem>
+<TabItem value="check-macos" label="MacOS" default>
```
jcsp, line 57: It's a bit odd to have MacOS instructions here -- afaik we do not support redpanda on MacOS, and the `brew` package is just rpk.  So there's no concept of a rolling upgrade there.

versioned_docs/version-22.3/manage/shared/rolling-upgrades/_restart-impact.mdx
jcsp: `redpanda_cpu_busy_seconds_total` is indeed the place to monitor CPU utilization.

versioned_docs/version-22.3/manage/shared/rolling-upgrades/_restart-impact.mdx
hcoyote: Yeah, something like that would be good.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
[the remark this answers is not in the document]
hcoyote: +1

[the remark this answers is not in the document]
hcoyote: I'm not sure what the right answer is for this. 

It's going to depend on availability of free space on the cluster. Cluster operators may have a good reason not to.

You could say something like:

> If you have topics with replication.factor=1, Redpanda recommends temporarily increasing that if you have sufficient free disk space. This can help limit outages for this topic during the rolling upgrade. This operation would need to be done ahead of the rolling restart/upgrade to ensure sufficient time for the data to replicate to other brokers.

versioned_docs/version-22.3/manage/shared/rolling-upgrades/_restart-impact.mdx
micheleRP: Thanks @jcsp and @RafalKorepta. Please see the updated list of metrics: https://deploy-preview-1246--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/rolling-upgrade/#check-metrics, and please confirm what they should look for with `redpanda_cpu_busy_seconds_total`. Should it be similar to what we have here:   https://docs.redpanda.com/docs/manage/monitoring/#infrastructure-resources

RafalKorepta: LGTM

RafalKorepta (approved):

versioned_docs/version-22.3/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: See https://github.com/redpanda-data/console/releases, and we cover MacOS in other sections of the doc. But I'm not on Mac, so I can't provide sample output. It'd be easier for maintaining this file if we don't include MacOS here. What do you think?

docs/manage/cluster-maintenance/rolling-upgrade.mdx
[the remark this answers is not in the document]
micheleRP: Thanks, @hcoyote! I updated the intro to include this.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
mattschumpert: The above is corrrect

mattschumpert: @hcoyote we don't support major version downgrade once the last node in the cluster has been upgraded (and the logical version in our metadata is bumped, new features get activated, etc.). We never have.  There's some interest to do so but not particularly strong.

mattschumpert:
@micheleRP 

I definitely like that this content has been combined, cleaned up, etc. Rolling upgrade is the base case of how customers want to upgrade so it certainly should be on the same page (our main upgrade page).

That said I still find it odd that the page isn't entitled 'Upgrading a Cluster' or better yet, just 'Upgrade/Upgrades'.  If I'm looking to 'Upgrade' thats what my eyes are scanning for.   'Rolling Upgrade' is a detail of how we suggest customers do an upgrade on a production cluster. But It's also perfectly possible to do a hard upgrade (non-rolling) of, for example, a dev cluster, where you really don't care if its rolling because nothing is running on it and you don't want to spend the time doing a rolling upgrade (especially if not on k8s where it kind of happens for you).

On this page, I'd expect to see the an overview explaining how upgrades work overall (e.g. upgrade isn't finalized until the last node is upgraded), the kernel of how to do upgrade a single node, and then explanation/example of rolling upgrade.  Instead this jumps straight into rolling upgrade, which feels slightly awkward.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
hcoyote: We need to call that out then, if the only safe way to move backwards is if we haven't finalized the version upgrade on all brokers. and I'd elevate it to an important or warning block if we can do that.

mattschumpert (approved):
Minor issue is that the 'restart brokers' step doesn't show the actual command to do so.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
daisukebe: @micheleRP, are you on it? This is still showing a wrong example output for Linux per https://deploy-preview-1246--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/rolling-upgrade/.

micheleRP: Hi @daisukebe! This issue of showing different output for docker/macOS/linux is the only remaining item on this PR. @hcoyote has a similar comment above. Can either of you can provide the output I should add for macOS & linux? If not, and if the sample docker output is confusing, I'd like to just remove this sample output... since it could be hard to maintain the file going forward for all 3 (and I also don't want to set the precedence that we're going to provide sample output for multiple OS in other files).

daisukebe: @micheleRP, thanks for tracking it!
Well, I'm with you. Given we don't support either MacOS or Docker based installations in a production use, it doesn't basically make sense to guide the users to do a rolling upgrade for those deployments. IIRC, John or Travic mentioned the same earlier in this PR.

@mattschumpert, should we leave the MacOS and Docker options as is in this doc?

docs/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: I've removed this example output. The doc team still needs to standardize on how much example output we include and the format for it. This example could be more confusing than helpful.

docs/manage/cluster-maintenance/rolling-upgrade.mdx
micheleRP: @daisukebe I'm going to leave these in for now, since users could possibly upgrade a dev cluster. The doc team still needs to standardize on how much example output we want to provide and the format for it. For now, I'm going to merge this PR. Please feel free to add a new doc issue if you want more updates to this page, or for issues about how we provide example output.

PR #1282, opened 2023-02-23
docs: Add note to include `expand` parameters in reverse-proxy rewrite rules

### LINKED ISSUE #774: Console>Kafka Connect - Add a note in case of reverse proxy to include "expand" parameters in reverse-proxy rewrite rules
### Describe the Issue
We should add a note for Redpanda Console >Kafka Connect where customers have a reverse proxy ..
they need to ensure any rewrite rules include URL parameters that are part of the "connectors" endpoint RP console sends to kafka-connect

We had a customer encounter an issue with this and there is a chance of there customers could hit the same issue...
so having a note in the docs on this would help these customers and avoid support tickets being raised ..

This is for redpanda folk implementing Redpanda Console , specifically with Kafka Connect 

https://docs.redpanda.com/docs/platform/console/features/kafka-connect/

have a blurb.. something like :

>  Redpanda Console make’s a request to the Kakfa Connect “connectors” endpoint with the “expand” parameter.      GET http://<kafka_connect_host>:8083/connectors?expand=info&expand=status
>  
>  If you have a reverse-proxy between Redpanda-Console and Kafka Connect, ensure the rewrite rules account for these parameters, otherwise the connector details will not be returned correctly

 

 
### If new feature, in which release is it included or expected? 
Not a new feature

### If the requested update is from customer feedback or a Community slack conversation, provide a link:
This comes from a support ticket.  
(585)
Also see slack
https://redpandadata.slack.com/archives/C02RQPX6A1K/p1667324388678719

JakeSCahill:
Review deadline: February 24

Fixes #774


Feediver1 (approved):
LGTM

Issue #958, opened 2022-12-19
Document topic properties


shorgi:
### Describe the Issue

Topic properties (compression.type, cleanup.policy, ...) are not documented in Redpanda docs.  

This leaves customers without guidance on how to configure topics, say through `rpk topic alter-config`. This also creates confusion between documented cluster properties and undocumented topic properties with similar names (for example, `retention_bytes` vs. `retention.bytes`).

### Updates to existing documentation

Add reference docs for topic properties.

Update existing pages with links to topic properties, including:

- https://docs.redpanda.com/docs/platform/data-management/tiered-storage/
- https://docs.redpanda.com/docs/platform/deployment/disk-utilization/#configure-message-retention

### Additional notes

See the supported and unsupported topic properties in https://github.com/redpanda-data/redpanda/blob/dev/src/v/kafka/server/handlers/topics/types.h

This is related to #853

shorgi:
Feedback from @eddieturizo:

> while consuming [Cluster Configuration](https://docs.redpanda.com/docs/manage/cluster-maintenance/configuration/), I came across:
Some properties can be configured at either the cluster level or the topic level, such as retention_bytes (cluster-level) and retention.bytes (topic-level). Topic-level properties override cluster-level properties.
i was expecting a link to topic level properties on that page, but it does not seem to exist.

PR #1290, opened 2023-02-24
Add topic properties reference

shorgi:
Review deadline: Wed Jun 14

Fixes:
- #958 

Preview:
- New [Topic configuration properties reference page](https://deploy-preview-1290--redpanda-documentation.netlify.app/docs/reference/topic-properties/)


docs/manage/cluster-maintenance/disk-utilization.mdx
```diff
@@ -31,9 +31,9 @@ Set retention properties at the topic level or the cluster level. If a value isn
 
 | Retention property | Cluster level | Topic level <br /> (overrides cluster configuration) |
 | ---               | ---                                                |  ---                                 |
-| Time-based        | [delete_retention_ms](../../../reference/cluster-properties#delete_retention_ms) | `retention.ms` <br /> No default     |
-| Size-based        | [retention_bytes](../../../reference/cluster-properties#retention_bytes) | `retention.bytes`  <br /> No default |
-| Segment size      | [log_segment_size](../../../reference/tunable-properties#log_segment_size) | N/A                                  |
+| Time-based        | [delete_retention_ms](../../../reference/cluster-properties#delete_retention_ms) | [retention.ms](../../../reference/topic-properties#retentionms)    |
+| Size-based        | [retention_bytes](../../../reference/cluster-properties#retention_bytes) | [retention.bytes](../../../reference/topic-properties#retentionbytes) |
+| Segment size      | [log_segment_size](../../../reference/tunable-properties#log_segment_size) | [segment.bytes](../../../reference/topic-properties#segmentbytes)                                |
```
micheleRP, line 36: Hi @shorgi. Could you please pick up the latest file for this? It now includes `retention.local.target.ms` and `retention.local.target.bytes`. I'd removed the links and kept defaults in there since we couldn't link to the topic properties yet, but with topic properties linkable now, we can remove the defaults from this table!

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,399 @@
...
+- [Tiered Storage](../../manage/tiered-storage)
+
+---
+
+#### redpanda.remote.recovery
+
+A flag that enables the recovery or reproduction of a topic from object storage for Tiered Storage. The recovered data is saved in local storage, and the maximum amount of recovered data is determined by the local storage retention limits of the topic.
+
+:::tip
+`redpanda.remote.recovery` must be configured at topic creation. It cannot be configured, and does not apply, for existing topics.
```
AnneLeeperRP, line 333: Change to active voice (and avoid "nominalization"):
```suggestion
You can only configure `redpanda.remote.recovery` when you create a topic. You cannot apply this setting to existing topics.
```

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,399 @@
...
+
+A flag that enables deletion of data from object storage for Tiered Storage when it's deleted from local storage for a topic.
+
+:::note
+`redpanda.remote.delete` doesn't apply to Remote Read Replica topics: a Remote Read Replica topic isn't deleted from object storage when this flag is `true`.
+:::
+
+**Default**:
+
+- `false` for topics created prior to Redpanda 22.3.
```
AnneLeeperRP, line 354:
```suggestion
- `false` for topics created using Redpanda 22.2 or earlier.
```

docs/reference/tunable-properties.mdx
```diff
@@ -1610,7 +1612,9 @@ The upper bound for the topic property `segment.ms` and the cluster property `lo
 
 ### log_segment_size
 
-Default size of log segment. Per topic, it is superseded by topic property `segment.bytes`.
+Default size of a log segment.
+
+`log_segment_size` is overridden per topic by the topic property [`segment.bytes`](../topic-properties#segmentbytes).
```
AnneLeeperRP, line 1617:
```suggestion
The topic property [`segment.bytes`](../topic-properties#segmentbytes) overrides `log_segment_size`.
```

docs/reference/tunable-properties.mdx
```diff
@@ -956,6 +956,8 @@ Number of partitions in the internal group membership topic.
 
 Maximum size of a batch processed by the server. If the batch is compressed, this limit applies to the compressed batch size.
 
+`kafka_batch_max_bytes` is overridden per topic by the topic property [`max.message.bytes`](../topic-properties#maxmessagebytes).
```
AnneLeeperRP, line 959:
```suggestion
The topic property [`max.message.bytes`](../topic-properties#maxmessagebytes) overrides `kafka_batch_max_bytes`.
```

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,399 @@
...
+    content="Configure topic-level properties for Redpanda that apply per topic and override corresponding node or cluster properties."
+  />
+</head>
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+A topic-level property sets a Redpanda or Kafka configuration for a particular topic. 
+
+## Property overrides
```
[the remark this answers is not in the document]
shorgi: I like including "override" because it's a commonly used and understood term for Kafka configuration. "Topic-level" is less used, so it doesn't have the immediate connotation of an override. Thinking about alternatives ...

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,399 @@
...
+
+Many topic-level properties override the value of a corresponding [cluster property](../../manage/cluster-maintenance/cluster-property-configuration). If a topic-level property isn't set, the corresponding cluster property takes effect by default.
```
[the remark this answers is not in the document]
shorgi: Yes, that's the gist. The "topic property at the cluster level" is the cluster property itself. I'll clarify this.

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,399 @@
...
+```bash
+kubectl exec <pod-name> -- rpk topic create -c redpanda.remote.recovery=true -c redpanda.remote.write=true -c redpanda.remote.read=true <topic-name>
+```
+
+</TabItem>
+</Tabs>
+
+### Modify topic properties
+
+To set or modify topic properties of an existing topic, use the [rpk topic alter-config](../rpk/rpk-topic/rpk-topic-alter-config) command.
```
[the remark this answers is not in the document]
shorgi: Set = modify a property of an existing topic that wasn't explicitly set upon topic creation. I see your point, it's redundant to "modify", will change this to "To modify topic properties of an existing topic ..."

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,399 @@
...
+
+### Disk space properties
+
+Configure properties to manage the disk space used by a topic:
+
+- Clean up log segments by deletion and/or compaction ([`cleanup.policy`](#cleanuppolicy)).
+- Retain logs up to a maximum size per partition before cleanup ([`retention.bytes`](#retentionbytes)).
+- Retain logs for a maximum duration before cleanup ([`retention.ms`](#retentionms)).
+- Periodically roll logs to inactivate segments ([`segment.ms`](#segmentms)).
+- Retain log segments up to a maximum size ([`segment.bytes`](#segmentbytes)).
```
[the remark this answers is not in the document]
shorgi: I prefer the bulleted list, because it emphasizes the summary descriptions (with the links to properties as convenient asides).

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,399 @@
...
+
+**Related topics**:
+
+- [Tiered Storage](../../manage/tiered-storage)
+
+---
+
+#### redpanda.remote.read
+
+A flag for enabling Redpanda to fetch data for a topic from object storage to local storage. When set `true` together with [redpanda.remote.write](#redpandaremotewrite), it enables the [Tiered Storage](../../manage/tiered-storage) feature.
```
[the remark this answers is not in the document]
shorgi: From https://vectorizedio.atlassian.net/wiki/spaces/DOC/pages/92635182/Redpanda+Documentation+Style+Guide#Redpanda-specific-word-list, use "Redpanda" to refer to the product.

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,399 @@
...
+
+**Related topics**:
+
+- [Tiered Storage](../../manage/tiered-storage)
+
+---
+
+#### retention.local.target.bytes
+
+A size-based retention limit that configures the maximum size that a topic partition in local storage of Tiered Storage can grow before becoming eligible for cleanup. It is equivalent to [retention.bytes](#retentionbytes) without Tiered Storage.
```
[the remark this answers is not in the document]
shorgi: It's the local storage of/for Tiered Storage. I'll reword it to be clearer.

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,392 @@
...
+</head>
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+A topic-level property sets a Redpanda or Kafka configuration for a particular topic. 
+
+Many [cluster properties](../../manage/cluster-maintenance/cluster-property-configuration) have corresponding topic-level properties. The value of a cluster property applies for all topics of a cluster. To customize the value for a topic, you can set a topic-level property that overrides the value of the corresponding cluster property.
+
+| Topic property                                    | Corresponding Cluster property                                                      |
```
AnneLeeperRP, line 22: How about switching the order (and retitling) so "Cluster property" is on the left and "Corresponding topic property" is on the right?
The lead-in sentence focusses on cluster properties, so that's the frame of reference.

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,392 @@
...
+
+### Disk space properties
+
+The following properties manage the disk space used by a topic:
+
+- Clean up log segments by deletion and/or compaction ([`cleanup.policy`](#cleanuppolicy)).
+- Retain logs up to a maximum size per partition before cleanup ([`retention.bytes`](#retentionbytes)).
+- Retain logs for a maximum duration before cleanup ([`retention.ms`](#retentionms)).
+- Periodically roll logs to inactivate segments ([`segment.ms`](#segmentms)).
+- Retain log segments up to a maximum size ([`segment.bytes`](#segmentbytes)).
```
AnneLeeperRP, line 117: Suggest listing <property name> - description (switch the order).
The lead-in sentence says "The following _properties_...", so properties should come first.

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,392 @@
...
+| [`retention.bytes`](#retentionbytes)              | [`retention_bytes`](../cluster-properties#retention_bytes)                       |
+| [`retention.ms`](#retentionms)                    | [`delete_retention_ms`](../cluster-properties#delete_retention_ms)               |
+| [`segment.ms`](#segmentms)                        | [`log_segment_ms`](../cluster-properties#log_segment_ms)                         |
+| [`segment.bytes`](#segmentbytes)                  | [`log_segment_size`](../tunable-properties#log_segment_size)                     |
+| [`compression.type`](#compressiontype)            | [`log_compression_type`](../cluster-properties#log_compression_type)             |
+| [`message.timestamp.type`](#messagetimestamptype) | [`log_message_timestamp_type`](../cluster-properties#log_message_timestamp_type) |
+| [`max.message.bytes`](#maxmessagebytes)           | [`kafka_batch_max_bytes`](../tunable-properties#kafka_batch_max_bytes)           |
+| [`replication.factor`](#replicationfactor)        | [`default_topic_replications`](../cluster-properties#default_topic_replications)   |
+
+## Examples
```
AnneLeeperRP, line 34: Let's move "Examples" into a tutorial. The tutorial can link to the reference topic as appropriate.

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,392 @@
...
+    content="Configure topic-level properties for Redpanda that override corresponding node or cluster properties."
+  />
+</head>
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+A topic-level property sets a Redpanda or Kafka configuration for a particular topic. 
+
+Many [cluster properties](../../manage/cluster-maintenance/cluster-property-configuration) have corresponding topic-level properties. The value of a cluster property applies for all topics of a cluster. To customize the value for a topic, you can set a topic-level property that overrides the value of the corresponding cluster property.
```
AnneLeeperRP, line 20:
```suggestion
Many [cluster properties](../../manage/cluster-maintenance/cluster-property-configuration) have corresponding topic-level properties. The value of a cluster property applies for all topics of a cluster. To customize the value for a topic, you can set a topic-level property on the command line that overrides the value of the corresponding cluster property.
```

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,392 @@
...
+
+**Default**: null
+
+**Related topics**:
+
+- [Configure message retention](../../manage/cluster-maintenance/disk-utilization#configure-message-retention)
+
+---
+
+#### segment.ms
```
daisukebe, line 173: Whereas it's mentioned in the linked page, `Log rolling`, it's better to mention about `log_segment_ms_min` and `log_segment_ms_max`, which configure the lower and the upper bound on the property, respectively. 

https://github.com/redpanda-data/redpanda/blob/v23.1.1/src/v/config/configuration.cc#L94-L111

docs/manage/cluster-maintenance/disk-utilization.mdx
```diff
@@ -117,12 +117,14 @@ rpk topic alter-config <topic> --set segment.bytes=<segment_size>
 
 ### Segment size for compacted segments
  
-Compaction saves space by retaining only the last known value for each message key within the segment. The latest value for each key is preserved, and all older values are discarded. 
+Compaction, or key-based retention, saves space by retaining only the last known value for each message key within the segment. The latest value for each key is preserved, and all older values are discarded.
```
mmaslankaprv, line 121: Here we may mention that this is a best effort, and there still may be duplicates

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,392 @@
...
+| [`retention.bytes`](#retentionbytes)              | [`retention_bytes`](../cluster-properties#retention_bytes)                       |
+| [`retention.ms`](#retentionms)                    | [`delete_retention_ms`](../cluster-properties#delete_retention_ms)               |
+| [`segment.ms`](#segmentms)                        | [`log_segment_ms`](../cluster-properties#log_segment_ms)                         |
+| [`segment.bytes`](#segmentbytes)                  | [`log_segment_size`](../tunable-properties#log_segment_size)                     |
+| [`compression.type`](#compressiontype)            | [`log_compression_type`](../cluster-properties#log_compression_type)             |
+| [`message.timestamp.type`](#messagetimestamptype) | [`log_message_timestamp_type`](../cluster-properties#log_message_timestamp_type) |
+| [`max.message.bytes`](#maxmessagebytes)           | [`kafka_batch_max_bytes`](../tunable-properties#kafka_batch_max_bytes)           |
+| [`replication.factor`](#replicationfactor)        | [`default_topic_replications`](../cluster-properties#default_topic_replications)   |
```
AnneLeeperRP, line 32: The `replication.factor` property was removed in an earlier release (21.11 I think?). Check this Slack convo:
https://redpandadata.slack.com/archives/C04LYNJQUGN/p1679346381485699?thread_ts=1679345262.700319&cid=C04LYNJQUGN

docs/reference/topic-properties.mdx
shorgi: Because users will more likely navigate to this page to find a reference for a topic property, this should be a lookup table from a topic property they're looking for to its corresponding cluster property.

docs/reference/topic-properties.mdx
shorgi: Thanks, will remove from this table, and will review whether it should be in this reference at all as a topic property.

docs/reference/topic-properties.mdx
shorgi: I prefer to change the lead-in sentence, because I intended for this bulleted list to a be a lookup "table" from a use-case to a property.

docs/reference/topic-properties.mdx
daisukebe: `replication.factor` is still a valid topic property but it's not just listed in the DescribeTopics response, Kafka protocol wise. As we can change it via `rpk topic alter-config [TOPICS...] --set replication.factor=3`, [ref](https://docs.redpanda.com/docs/manage/data-migration/#change-topic-replication-factor), it's better to leave it here. @mmaslankaprv @twmb , thoughts?

shorgi: My latest commits ([9c6e636](https://github.com/redpanda-data/documentation/pull/1290/commits/9c6e636962640a08b0e2862618dc787a41149e32), [2c94b93](https://github.com/redpanda-data/documentation/pull/1290/commits/2c94b9361e988cdb47eaa07d48ef7f8adf1adc8b)) added clarification about `replication.factor` being Redpanda-specific and configurable via `rpk topic alter-config`.

docs/develop/produce-data/configure-producers.mdx
```diff
@@ -220,7 +220,7 @@ high chance messages arrive around the same time to be batched together.
 
 #### `compression.type`
 
-`compression.type` controls how the producer should compress a batch of messages
+[compression.type](../../../reference/topic-properties#compressiontype) controls how the producer should compress a batch of messages
 before sending it to the broker. The default is `none`, which means the batch of
```
weeco, line 224: I think this document should note that the entire documentation refers to the Java Kafka client (of version x?). The defaults and configuration names are different between all Kafka clients. E.g our commonly recommended Kafka client franz-go indeed has a different compression type. 

Also the Java client's defaults change between version. Most recently a couple changes were introduced with version 3 of the Java client. I could not see any hints about this in configure-producers.mdx doc.

docs/develop/transactions.mdx
```diff
@@ -258,7 +258,7 @@ Different transactions require different approaches to handling failures within
 
 ## Transactions with compacted segments
 
-Transactions are supported on topics with compaction configured. When [`log_cleanup_policy`](../../reference/cluster-properties/#log_cleanup_policy) is set to `compact`, the compaction process removes the aborted transaction's data and all transactional control markers from the log. The resulting compacted segment contains only committed data batches (and potentially harmless gaps in the offsets due to skipped batches). 
+Transactions are supported on topics with compaction configured. When [`log_cleanup_policy`](../../reference/cluster-properties/#log_cleanup_policy) or [`cleanup.policy`](../../reference/topic-properties#cleanuppolicy) is set with `compact`, the compaction process removes the aborted transaction's data and all transactional control markers from the log. The resulting compacted segment contains only committed data batches (and potentially harmless gaps in the offsets due to skipped batches).
```
weeco, line 261: How about the cleanup.policy `compact,delete` ?

docs/manage/cluster-maintenance/disk-utilization.mdx
```diff
@@ -117,12 +117,14 @@ rpk topic alter-config <topic> --set segment.bytes=<segment_size>
 
 ### Segment size for compacted segments
```
weeco, line 118: Did you mean to write "compacted topics" instead of "compacted segments" here?

docs/manage/cluster-maintenance/disk-utilization.mdx
```diff
@@ -117,12 +117,14 @@ rpk topic alter-config <topic> --set segment.bytes=<segment_size>
...
 
 Setting a `segment.bytes` size on a topic applies whether the topic is compacted or not, and the `max_compacted_log_segment_size` property applies to compacted topics regardless of any other properties. The `max_compacted_log_segment_size` property controls how many segments are merged together. For example, if you set `segment.bytes` to 128 MB, but leave `max_compacted_log_segment_size` at 5 GB, then you get 128 MB segments when they're written, but up to 5 GB segments after compaction.
 
+Redpanda does compaction periodically in the background. The compaction period is configured by the cluster property [log_compaction_interval_ms](../../../reference/cluster-properties#log_compaction_interval_ms).
+
 Keep in mind that very large segments prevent compaction, and very small segments increase the risk of encountering resource limits. To calculate an upper limit on segment size, divide the disk size by the number of partitions. For example, if you have a 128 GB disk and 1000 partitions, the upper limit of the segment size is `134217728`. Default is `1073741824`.
```
weeco, line 128: > Keep in mind that very large segments prevent compaction

I think this is important to consider, on a first glance it seemed that this section however does not explain why that is right? In case I have not just missed it I could imagine linking to the place where the compaction process is explained (segments which are eligible for compaction specifically).

docs/reference/cluster-properties.mdx
```diff
@@ -946,6 +948,14 @@ Raft learner recovery rate limit. Throttles the rate of data communicated to nod
 
 Delete segments that are older than this age.
 
+The topic property [`retention.ms`](../topic-properties#retentionms) overrides the value of `delete_retention_ms` at the topic level.
+
+:::note
+The `delete_retention_ms` cluster property is different than the `delete.retention.ms` Kafka topic property.
```
weeco, line 954:
```suggestion
The `delete_retention_ms` cluster property is different than the `delete.retention.ms` Apache Kafka topic property.
```

docs/reference/topic-properties.mdx
twmb: @daisukebe I thought we ignored replication.factor being changed now. I'm not sure why we don't reject that as a config option. I'd like to wait for @mmaslankaprv 's thoughts here too

daisukebe: Hmm ohk. It's still available.
```
daisuke@/playground/ $ rpk topic describe -p test --brokers localhost:9093
PARTITION  LEADER  EPOCH  REPLICAS  LOG-START-OFFSET  HIGH-WATERMARK
0          1       1      [1 2 3]   0                 0
daisuke@/playground/ $ rpk topic alter-config test -s replication.factor=1 --brokers localhost:9093
TOPIC  STATUS
test   OK
daisuke@/playground/ $ rpk topic describe -p test --brokers localhost:9093
PARTITION  LEADER  EPOCH  REPLICAS  LOG-START-OFFSET  HIGH-WATERMARK
0          1       1      [1]       0                 0
```
```
daisuke@/playground/ $ rpk redpanda admin brokers list --hosts localhost:9645
NODE-ID  NUM-CORES  MEMBERSHIP-STATUS  IS-ALIVE  BROKER-VERSION
1        2          active             true      v23.1.2 - 6002d4db8cefabafb155030bc18321afe3fc2db3
2        2          active             true      v23.1.2 - 6002d4db8cefabafb155030bc18321afe3fc2db3
3        2          active             true      v23.1.2 - 6002d4db8cefabafb155030bc18321afe3fc2db3
```

shorgi: @mmaslankaprv @twmb @daisukebe - I'd like to resolve this comment by the end of day Thursday Apr 6. If it's still unresolved, I propose to remove `replication.factor` for now, and to review its inclusion in a separate PR.

docs/develop/produce-data/configure-producers.mdx
shorgi: (Aside: configure-producers is being revised in https://github.com/redpanda-data/documentation/pull/1448)

docs/manage/cluster-maintenance/disk-utilization.mdx
shorgi: @weeco and @micheleRP (I think you updated this last?) - I think large segments don't always prevent compaction but slow it down. See suggestion:

```suggestion
Keep in mind that very large segments delay, or possibly prevent, compaction. A very large active segment cannot be cleaned up or compacted until it is closed, and very large closed segments require significant memory and CPU to process for compaction. Very small segments increase the frequency of processing for applying compaction and resource limits. To calculate an upper limit on segment size, divide the disk size by the number of partitions. For example, if you have a 128 GB disk and 1000 partitions, the upper limit of the segment size is `134217728`. Default is `1073741824`. 
```

Also, the docs are missing an explanation of the compaction concept (and other Kafka topic concepts). That'll be added in a separate PR and linked to here.

weeco: That's a good addition I think. After reading this the only question I'd still have - what's the problem with frequent processing of very small segments (and what does very small mean here - like one segment per minute or more segments)? Seems fine to me to have compact "frequently" or are we saying that this ist just way more inefficient rather than finding a reasonable sized segment?

docs/reference/topic-properties.mdx
mmaslankaprv: Replication factor can be changed by alter configuration, it is not reported as it isn't a valid Kafka property. So it only works one way. Also when altered it isn't simply setting the property but actually changing replica sets of topic partitions.

docs/manage/cluster-maintenance/disk-utilization.mdx
```diff
@@ -116,15 +116,17 @@ To configure log segment size on a topic:
 rpk topic alter-config <topic> --set segment.bytes=<segment_size>
 ```
 
-### Segment size for compacted segments
+### Segment size for compacted topics
  
-Compaction saves space by retaining only the last known value for each message key within the segment. The latest value for each key is preserved, and all older values are discarded. 
+Compaction, or key-based retention, saves space by retaining at least the most recent value for a message key within a topic partition's log and discarding older values. Compaction runs periodically in the background in a best effort fashion, and it doesn't guarantee that there are no duplicate values per key.
```
Feediver1, line 121:
```suggestion
Compaction, or key-based retention, saves space by retaining at least the most recent value for a message key within a topic partition's log and discarding older values. Compaction runs periodically in the background in a best effort fashion, and doesn't guarantee that there are no duplicate values per key.
```

docs/manage/cluster-maintenance/disk-utilization.mdx
```diff
@@ -116,15 +116,17 @@ To configure log segment size on a topic:
...
 
 Setting a `segment.bytes` size on a topic applies whether the topic is compacted or not, and the `max_compacted_log_segment_size` property applies to compacted topics regardless of any other properties. The `max_compacted_log_segment_size` property controls how many segments are merged together. For example, if you set `segment.bytes` to 128 MB, but leave `max_compacted_log_segment_size` at 5 GB, then you get 128 MB segments when they're written, but up to 5 GB segments after compaction.
 
-Keep in mind that very large segments prevent compaction, and very small segments increase the risk of encountering resource limits. To calculate an upper limit on segment size, divide the disk size by the number of partitions. For example, if you have a 128 GB disk and 1000 partitions, the upper limit of the segment size is `134217728`. Default is `1073741824`. 
+Redpanda does compaction periodically in the background. The compaction period is configured by the cluster property [log_compaction_interval_ms](../../../reference/cluster-properties#log_compaction_interval_ms).
```
Feediver1, line 127:
```suggestion
Redpanda periodically performs compaction in the background. The compaction period is configured by the cluster property [log_compaction_interval_ms](../../../reference/cluster-properties#log_compaction_interval_ms).
```

docs/reference/cluster-properties.mdx
```diff
@@ -1041,6 +1043,14 @@ Raft learner recovery rate limit. Throttles the rate of data communicated to nod
 
 Delete segments that are older than this age.
 
+The topic property [`retention.ms`](../topic-properties#retentionms) overrides the value of `delete_retention_ms` at the topic level.
+
+:::note
```
Feediver1, line 1048: I'd make the fact that we don't support `delete.retention.ms` more prominent, and put in the introductory statement here--not at the end of the 2nd bullet.

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,419 @@
...
+  <meta
+    name="description"
+    content="Look up reference descriptions and examples about topic-level properties for Redpanda."
+  />
+</head>
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+Look up reference descriptions and examples of topic-level properties supported by Redpanda.
```
Feediver1, line 18: This intro seems inconsistent with the first statement in the other properties. None of those have the "About" section either. I think all of the property files should have a consistent intro. Maybe start with: 

Topic-level properties set the Redpanda or Kafka configuration for a particular topic. 

And then remove the About heading?  Concerned about consistency with the other properties ref intros.

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,419 @@
...
+| [`segment.bytes`](#segmentbytes)                  | [`log_segment_size`](../tunable-properties#log_segment_size)                     |
+| [`compression.type`](#compressiontype)            | [`log_compression_type`](../cluster-properties#log_compression_type)             |
+| [`message.timestamp.type`](#messagetimestamptype) | [`log_message_timestamp_type`](../cluster-properties#log_message_timestamp_type) |
+| [`max.message.bytes`](#maxmessagebytes)           | [`kafka_batch_max_bytes`](../tunable-properties#kafka_batch_max_bytes)           |
+| [`replication.factor`](#replicationfactor)        | [`default_topic_replications`](../cluster-properties#default_topic_replications)   |
+
+:::note
+The `SOURCE` output of the [`rpk topic describe <topic>`](../rpk/rpk-topic/rpk-topic-describe) command describes how the property was set for the topic:
+- `DEFAULT_CONFIG` was set by a Redpanda default.
+- `DYNAMIC_TOPIC_CONFIG` was set by the user specifically for the topic and overrides inherited default configurations, such as a default or a cluster-level property.
```
Feediver1, line 41:
```suggestion
The `SOURCE` output of the [`rpk topic describe <topic>`](../rpk/rpk-topic/rpk-topic-describe) command describes how the property is set for the topic:
- `DEFAULT_CONFIG` is set by a Redpanda default.
- `DYNAMIC_TOPIC_CONFIG` is set by the user specifically for the topic and overrides inherited default configurations, such as a default or a cluster-level property.
```

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,419 @@
...
+
+While [`rpk topic describe`](../rpk/rpk-topic/rpk-topic-describe) doesn't report `replication.factor` as a configuration, `replication.factor` can indeed be set via the [`rpk topic alter-config`](../rpk/rpk-topic/rpk-topic-alter-config) command.
```
Feediver1, line 43: Style Guide instructs not to use "via":

While [`rpk topic describe`](../rpk/rpk-topic/rpk-topic-describe) doesn't report `replication.factor` as a configuration, `replication.factor` can indeed be set using the [`rpk topic alter-config`](../rpk/rpk-topic/rpk-topic-alter-config) command.

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,419 @@
...
+#### segment.ms
+
+The maximum duration that a log segment of a topic is active (open for writes and not deletable). A periodic event, with `segment.ms` as its period, forcibly closes the active segment and transitions, or rolls, to a new active segment. The closed (inactive) segment is then eligible to be cleaned up according to cleanup and retention properties.
+
+If set to a positive duration, `segment.ms` overrides the cluster property [`log_segment_ms`](../cluster-properties#log_segment_ms) and its lower and upper bounds set by [`log_segment_ms_min`](../cluster-properties#log_segment_ms_min) and [`log_segment_ms_max`](../cluster-properties#log_segment_ms_max), respectively.
+
+**Default**: null
+
+**Related topics**:
+[Log rolling](../../manage/cluster-maintenance/disk-utilization#log-rolling)
```
Feediver1, line 194: Other single related topic entries are in bullets (see preceding property). This one is not. Be consistent with how you present.

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,419 @@
...
+The source of a message's timestamp: either the message's creation time or its log append time.
+
+When `message.timestamp.type` is set, it overrides the cluster property [`log_message_timestamp_type`](../cluster-properties#log_message_timestamp_type) for the topic.
+
+**Default**: `'CreateTime'`
+
+**Values**:
+
+- `'CreateTime'`
+- `'LogAppendTime'`
```
Feediver1, line 263: Are the backticks supposed to appear here?

```suggestion
- `CreateTime`
- `LogAppendTime`
```

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,419 @@
...
+
+---
+
+#### message.timestamp.type
+
+The source of a message's timestamp: either the message's creation time or its log append time.
+
+When `message.timestamp.type` is set, it overrides the cluster property [`log_message_timestamp_type`](../cluster-properties#log_message_timestamp_type) for the topic.
+
+**Default**: `'CreateTime'`
```
Feediver1, line 258:
```suggestion
**Default**: `CreateTime`
```

docs/reference/topic-properties.mdx
```diff
@@ -0,0 +1,419 @@
...
+Configure Redpanda-specific topic properties.
+
+#### replication.factor
+
+The number of replicas of a topic to save in different nodes (brokers) of a cluster.
+
+If `replication.factor` is set to a positive value, it overrides the cluster property [default_topic_replications](../cluster-properties#default_topic_replications) for the topic.
+
+:::note
+Although `replication.factor` isn't returned or displayed by [`rpk topic describe`](../rpk/rpk-topic/rpk-topic-describe) as a valid Kafka property, it indeed can be set by [`rpk topic alter-config`](../rpk/rpk-topic/rpk-topic-alter-config). When the `replication.factor` of a topic is altered, it isn't simply a property value that's updated, but rather the actual replica sets of topic partitions are changed.
```
Feediver1, line 404:
```suggestion
Although `replication.factor` isn't returned or displayed by [`rpk topic describe`](../rpk/rpk-topic/rpk-topic-describe) as a valid Kafka property, you can set it using [`rpk topic alter-config`](../rpk/rpk-topic/rpk-topic-alter-config). When the `replication.factor` of a topic is altered, it isn't simply a property value that's updated, but rather the actual replica sets of topic partitions that are changed.
```

Feediver1 (approved):
See suggestions. Backtick formatting needs to be fixed, and anything to do with consistency. Revised one or two items because I had to read them over a few times--figured if I had to, users might too.

docs/reference/topic-properties.mdx
shorgi: All the properties references will be updated to have more substantial conceptual "About" sections in upcoming PRs. Before updating them all, for consistency this "About ..." heading will be removed, but its content will be retained.

shorgi:
This PR will be merged EOD, Friday June 23

PR #1352, opened 2023-03-08
Fix placeholders

JakeSCahill:
This PR fixes a bug in the recent editable placeholders UX feature where wrapped quotes were removed and multiple placeholders in a single `<span>` were replaced with a single placeholder.

We now properly handle cases where one or more placeholders appear in a single `<span>` element.

We also remove some complexity from the `<CopyButton/>` component by using the `::after` pseudo-element to insert the cursors instead of inserting them into `<span>` elements.

Before:

After:

Preview: https://deploy-preview-1352--redpanda-documentation.netlify.app/docs/manage/kubernetes/security/sasl-kubernetes/


src/theme/CodeBlock/CopyButton/index.js
```diff
@@ -1,7 +1,7 @@
 import React, {useState, useRef, useEffect} from 'react';
 import clsx from 'clsx';
 // @ts-expect-error: TODO, we need to make theme-classic have type: module
-import copy from 'copy-text-to-clipboard';
+//import copy from 'copy-text-to-clipboard';
```
Deflaimun, line 4: Remove unused code

Deflaimun (approved):
Just check the previous comment, but otherwise LGTM.

Deflaimun:
Unrelated to the current issue. When editing a code block with long strings it can generate an overflow bar. While this isn't bad per se, it should instead wrap the content and break into the next line

image

You can use this website to generate long strings to test this behavior.
http://www.unit-conversion.info/texttools/random-string-generator/

JakeSCahill:
Thanks @Deflaimun - since this bug is currently breaking some of our code snippets, I'll merge this one first and look into preventing scrolling in another PR 😄

PR #1359, opened 2023-03-10
Use `latest` connectors image version in owl-shop example

bochenekmartin:
The version `redpandadata/connectors:1.0.0-dev-dff1c57` used in the doc does not exist so the container from the example does not work.

### DIFF
--- a/versioned_docs/version-22.3/reference/docker-compose.mdx
+++ b/versioned_docs/version-22.3/reference/docker-compose.mdx
@@ -105,7 +105,7 @@ services:
       - redpanda
 
   connect:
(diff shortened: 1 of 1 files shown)


Deflaimun:
@bochenekmartin should we replace all of our versions to `latest` ?
For example 
From 
```yaml
    image: docker.redpanda.com/redpandadata/redpanda:v22.3.11
```

to 
```yaml
    image: docker.redpanda.com/redpandadata/redpanda:latest

micheleRP (approved):
Not sure if other instances should also change to `latest `(like on 23.1 page), but approving for this 22.3 fix. Thank you!

bochenekmartin:
The change applies only to connectors. redpanda or console should not be changed to `latest`.

PR #1376, opened 2023-03-14
deployment: Add a paragraph about safely doing back to back upgrades

jcsp:
This is aimed at folks writing installers.

Preview:
- https://deploy-preview-1376--redpanda-documentation.netlify.app/docs/deploy/deployment-option/self-hosted/manual/production/production-deployment/#upgrade-considerations


micheleRP:
The link in the Upgrade considerations section of the Deploy for Production page goes to the Node Maintenance Mode page, but I think it'd be better to have one to the [Upgrade page](https://docs.redpanda.com/docs/manage/cluster-maintenance/rolling-upgrade/). 
@jcsp do you want me to add a similar line about doing back-to-back upgrades in #1246?

docs/deploy/deployment-option/self-hosted/manual/production/production-deployment.mdx
```diff
@@ -304,6 +304,13 @@ risk during upgrades. Check the cluster's health after upgrading each node. Roll
 back an upgrade to an earlier feature release is only supported until the last node
 has been updated, so it's important to identify any issues before that point.
 
+If doing "back to back" upgrades that step through several feature versions of
+Redpanda, then check for each upgrade to fully complete before proceeding to the
+next, by reading the `/v1/features` admin API endpoint and checking that `cluster_version`
+has increased.  Since Redpanda v23.1, this endpoint also includes a `node_latest_version`
+attribute, and installers may check that the cluster has activated any new functionality
+from a previous upgrade by checking for `cluster_version` == `node_latest_version`.
+
```
shorgi, line 313:
```suggestion
If upgrading multiple [feature release versions of Redpanda](../../../../../../manage/kubernetes/rolling-upgrade#find-a-new-version) in succession, make sure to verify that each version upgrades to completion before proceeding to the next version. You can verify by reading the `/v1/features` Admin API endpoint and checking that `cluster_version` has increased. 

  - With Redpanda 23.1 and later, the `/v1/features` endpoint also includes a `node_latest_version` attribute, and installers can verify that the cluster has activated any new functionality from a previous upgrade by checking for `cluster_version` == `node_latest_version`.
```

shorgi (commented):
What versions does checking for increased `cluster_version` apply, 22.x? If so, need to backport the first sentence of this change.

shorgi:
> The link in the Upgrade considerations section of the Deploy for Production page goes to the Node Maintenance Mode page, but I think it'd be better to have one to the [Upgrade page](https://docs.redpanda.com/docs/manage/cluster-maintenance/rolling-upgrade/). @jcsp do you want me to add a similar line about doing back-to-back upgrades in #1246?

@micheleRP - once @jcsp approves the edited content, let's merge this PR and create a follow-up PR to single source this.

jcsp:
>@jcsp do you want me to add a similar line about doing back-to-back upgrades in https://github.com/redpanda-data/documentation/pull/1246?

Probably not necessary.  It only takes seconds for Redpanda to register its new cluster version, so this kind of check should only be needed when someone is writing automation that might do updates in quick succession.

Although it would be ideal to check this kind of thing prior to an upgrade, I don't think we can reasonably ask people to check this by hand until we wrap up upgrades with a friendlier command for checking if the system is good to go (https://github.com/redpanda-data/planning/issues/19#issuecomment-1468731644)

shorgi (approved):

PR #1404, opened 2023-03-18
docs: #641 add info about cluster balancing

micheleRP:
resolves #641 

preview:


docs/manage/cluster-maintenance/cluster-balancing.mdx
```diff
@@ -15,18 +15,28 @@ Cluster balancing protects you from unbalanced systems that saturate resources o
...
 
-To manually change leadership, use the Admin API:
+Leadership balancing is enabled by default with the [`enable_leader_balancer`](../../../reference/cluster-properties/#enable_leader_balancer) property. Redpanda tries to distribute the entire set of leaders evenly across all shard/broker pairs. Automatic partition leadership balancing improves cluster performance by transferring the leadership of a broker's partitions to other replicas. It changes where data is read from and written to first, but leadership transfer does not involve any data transfer.
```
dotnwat, line 20: > but leadership transfer does not involve any data transfer. 

exactly. no replicas are moved

dotnwat: > improves cluster performance

yep. it help avoid hot spots.

docs/manage/cluster-maintenance/cluster-balancing.mdx
```diff
@@ -15,18 +15,28 @@ Cluster balancing protects you from unbalanced systems that saturate resources o
...
 
-To manually change leadership, use the Admin API:
+Leadership balancing is enabled by default with the [`enable_leader_balancer`](../../../reference/cluster-properties/#enable_leader_balancer) property. Redpanda tries to distribute the entire set of leaders evenly across all shard/broker pairs. Automatic partition leadership balancing improves cluster performance by transferring the leadership of a broker's partitions to other replicas. It changes where data is read from and written to first, but leadership transfer does not involve any data transfer. 
+
+:::note
+In addition to the periodic heartbeat, leadership balancing can also occur when the cluster leader changes (such as when a controller partition changes leader), or when a broker restarts. For more information, see [Impact of broker restarts](../../../manage/cluster-maintenance/rolling-upgrade/#impact-of-broker-restarts).
```
dotnwat, line 23: > leadership balancing can also occur when the cluster leader changes (such as when a controller partition changes leader), or when a broker restarts.

do we use the terminology "cluster leader"? i wouldn't expect here that we'd allude to the "controller partition leader" being an example of the cluster leader so much as they are equivalent.

> or when a broker restarts

things may have changed, but i don't recall a broker restart triggering a rebalance. leadership will move when a broker is down (e.g. because it is restarting), but rebalancing last time i checked was a periodic process. maybe @ballard26 can chime in here?

docs/manage/cluster-maintenance/cluster-balancing.mdx
```diff
@@ -35,16 +45,19 @@ In Kubernetes, run the `transfer_leadership` request on the Pod that is running
 
 ## Redpanda partition balancing
 
-Partition balancing moves data from most-loaded brokers to least-loaded brokers to ensure steady-state performance. It ensures predictable and stable performance across all brokers in a cluster. Partition balancing is invoked periodically, determined by the `partition_autobalancing_tick_interval_ms` property. Default is 30 seconds.
+While *leadership balancing* doesn't move any data, *partition balancing* does move partition replicas from most-loaded brokers to least-loaded brokers. Depending on the amount of data being transferred, this may take some time. Partition balancing is invoked periodically, determined by the [`partition_autobalancing_tick_interval_ms`](../../../reference/tunable-properties/#partition_autobalancing_tick_interval_ms) property.
```
dotnwat, line 48: > most-loaded brokers to least-loaded brokers

I wonder if we should be more opaque about the mechanism and instead talks about the goal of the balancing policy / heuristic like balancing load where load is measured as X. maybe @ztlpn or @mmaslankaprv could chime in here?

docs/manage/cluster-maintenance/cluster-balancing.mdx
```diff
@@ -35,16 +45,19 @@ In Kubernetes, run the `transfer_leadership` request on the Pod that is running
...
+While *leadership balancing* doesn't move any data, *partition balancing* does move partition replicas from most-loaded brokers to least-loaded brokers. Depending on the amount of data being transferred, this may take some time. Partition balancing is invoked periodically, determined by the [`partition_autobalancing_tick_interval_ms`](../../../reference/tunable-properties/#partition_autobalancing_tick_interval_ms) property. 
+
+Redpanda's default partition balancing ensures predictable and stable performance across all brokers in a cluster. It includes the following: 
 
-By default, Redpanda rebalances partition distribution when brokers are added or decommissioned. Continuous Data Balancing additionally rebalances partitions when brokers become unavailable or when disk space usage exceeds a threshold. 
+- When a broker is added to the cluster, some of the replicas are moved to that broker.
```
dotnwat, line 52: > some of the replicas are moved to that broker

yep. redpanda will take advantage of the additional cluster capacity.

docs/manage/cluster-maintenance/cluster-balancing.mdx
```diff
@@ -35,16 +45,19 @@ In Kubernetes, run the `transfer_leadership` request on the Pod that is running
...
+- When a broker is down for a configured timeout, replicas from the offline broker are moved to other brokers.
```
dotnwat, line 53: > replicas from the offline broker are moved to other brokers.

a pedantic reader may think "how can data be moved from an offline broker". mechanically what is happening here is that existing online replicas are used to construct a replacement replica copy on a new broker.

micheleRP: Thanks @dotnwat! I've updated this.

docs/manage/cluster-maintenance/cluster-balancing.mdx
mmaslankaprv: yeah, we shouldn't be very specific here as we do not move the data from least loaded broker to most loaded one. The only situation we move data is when broker is down or out of the disk space.

mattschumpert: "partition balancing does move partition replicas as needed to alleviate disk pressure and to ensure the configured replication factor across nodes and additional redundancy across failure domains such as racks is honored"

mattschumpert (approved):
Left a few comments but approving now anyways.Comment about 'cluster leader' needs to be addressed I think

docs/manage/cluster-maintenance/cluster-balancing.mdx
micheleRP: @dotnwat, @ballard26: Would you please review and verify the note in [this section](https://deploy-preview-1404--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/cluster-balancing/#partition-leadership-balancing). Is the description of the difference between the controller leader and the partition leader correct? I don't think we have a definition of the controller leader anywhere else in our doc. Please also verify the the impact of broker restarts. That linked section was recently added to the upgrade page. 

In addition to the periodic heartbeat, leadership balancing can also occur when a [broker restarts](https://deploy-preview-1404--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/rolling-upgrade/#impact-of-broker-restarts) or when the controller leader changes (such as when a controller partition changes leader). The controller leader manages the entire cluster. For example, when a broker is decommissioned, the controller leader creates a reallocation plan for all partition replicas allocated to that broker. The partition leader then handles the reconfiguration for its Raft group.

shorgi (approved):
See comments, approved

micheleRP:
> See comments, approved

Thanks for the nice improvements, @shorgi!

micheleRP:
@dotnwat, @mmaslankaprv, @ballard26 or @piyushredpanda: Could someone from engineering please approve this PR?

mmaslankaprv (approved):

PR #1412, opened 2023-03-21
docs: Document the decommissioning procedure

### LINKED ISSUE #1406: Document the process for decommissioning brokers in Kubernetes
### Describe the Issue

We need a document that describes the procedure for decommissioning brokers in a cluster.

Users may want to decommission brokers for a variety of reasons including when a pod loses its storage and the node_id associated with that broker is lost. When the pod comes back without storage, that broker is assigned a new node_id, so you must decommission the broker that was running inside the Pod that lost its storage.

Resources:
https://vectorizedio.atlassian.net/wiki/spaces/CORE/pages/346423369/Node+decommissioning

JakeSCahill:
Review deadline: March 27

Fixes #1406 

When the content is approved, I'll single-source it so that we have a corresponding doc for other deployments that doesn't include the `kubectl` commands or references to worker nodes.


RafalKorepta (approved):

docs/manage/decommission-brokers.mdx
```diff
@@ -0,0 +1,95 @@
...
+- You are replacing a node, for example to upgrade the Linux kernel or to replace the hardware.
+- You are removing a broker to decrease the size of the cluster.
+
+:::info
+When a broker is decommissioned, it cannot rejoin the cluster. If a broker with the same ID tries to rejoin the cluster, it is rejected.
+:::
+
+## What happens when a broker is decommissioned
+
+When a broker is decommissioned, the controller leader creates a reallocation plan for all partition replicas that are allocated to that broker. By default, this reallocation is done in batches of 50 to avoid overwhelming the remaining brokers with raft recovery. See [`partition_autobalancing_concurrent_moves`](../../reference/tunable-properties#partition_autobalancing_concurrent_moves).
```
micheleRP, line 35: not positive, but I think Raft should always be capitalized

docs/manage/kubernetes/decommission-brokers.mdx
```diff
@@ -0,0 +1,158 @@
...
+- You are replacing a worker node, for example to upgrade the Kubernetes cluster or to replace the hardware.
+- You are removing a broker to decrease the size of the cluster.
+
+:::info
+When a broker is decommissioned, it cannot rejoin the cluster. If a broker with the same ID tries to rejoin the cluster, it is rejected.
+:::
+
+## What happens when a broker is decommissioned
+
+When a broker is decommissioned, the controller leader creates a reallocation plan for all partition replicas that are allocated to that broker. By default, this reallocation is done in batches of 50 to avoid overwhelming the remaining brokers with raft recovery. See [`partition_autobalancing_concurrent_moves`](../../../reference/tunable-properties#partition_autobalancing_concurrent_moves).
```
micheleRP, line 41: not positive, but I think Raft should always be capitalized

micheleRP:
@JakeSCahill for the Linux version: this is located in Manage/Decommission Brokers. Do you think this should move to Manage/Cluster Maintenance/Decommission Brokers? (Same with Node Maintenance Mode--does that make more sense nested in Cluster Maintenance?)

docs/manage/decommission-brokers.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+The reallocation of each partition is translated into a Raft group reconfiguration and executed by the controller leader. The partition leader then handles the reconfiguration for its Raft group. Once the reallocation for a partition is complete, it is recorded in the controller log and the status is updated in the topic tables of each broker.
```
micheleRP, line 37: This is the first time I've read text about Redpanda's controller leader, and I don't understand difference between the controller leader and the partition leader (or controller leadership transfers and partition leadership transfers). Can you add some explanation here? Or maybe the Cluster Balancing page is a better place. Or glossary.

docs/manage/kubernetes/decommission-brokers.mdx
```diff
@@ -0,0 +1,158 @@
+---
+title: Decommission Brokers in Kubernetes
```
micheleRP, line 2: @JakeSCahill what do you think about having different file names for these 2 versions; e.g., `decommission-brokers.mdx` and `k-decommission-brokers.mdx`

docs/manage/decommission-brokers.mdx
JakeSCahill: @jcsp correct me if I'm wrong, but I believe the difference is:

The controller leader manages the entire cluster, while the partition leader is responsible for managing a subset of data (partition) within the cluster.

docs/manage/kubernetes/decommission-brokers.mdx
JakeSCahill: sure, we can do that. It might look weird in the URL though. Is the only reason so that we can easily differentiate the pages while authoring?

The page is already nested within a `kubernetes` folder: `docs/manage/kubernetes/decommission-brokers.mdx`

micheleRP: Yes, just for authoring. Probably a separate issue from this PR, but we should create some standard for naming; e.g., I noticed in your other PR that similarly adds a separate file for k8s, the original file is named `monitoring.mdx` and the new k8s file is named `monitor.mdx`. When we need to update 2 similar files (Linux/Kubernetes), it'll be easier if we can tell their difference by the file name. But we should follow some consistent naming.

docs/manage/cluster-maintenance/decommission-brokers.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+:::info
+When a broker is decommissioned, it cannot rejoin the cluster. If a broker with the same ID tries to rejoin the cluster, it is rejected.
+:::
+
+## What happens when a broker is decommissioned
+
+When a broker is decommissioned, the controller leader creates a reallocation plan for all partition replicas that are allocated to that broker. By default, this reallocation is done in batches of 50 to avoid overwhelming the remaining brokers with Raft recovery. See [`partition_autobalancing_concurrent_moves`](../../../reference/tunable-properties#partition_autobalancing_concurrent_moves).
+
+The reallocation of each partition is translated into a Raft group reconfiguration and executed by the controller leader. The partition leader then handles the reconfiguration for its Raft group. Once the reallocation for a partition is complete, it is recorded in the controller log and the status is updated in the topic tables of each broker.
```
micheleRP, line 37:
```suggestion
The reallocation of each partition is translated into a Raft group reconfiguration and executed by the controller leader. The partition leader then handles the reconfiguration for its Raft group. After the reallocation for a partition is complete, it is recorded in the controller log and the status is updated in the topic tables of each broker.
```

docs/manage/cluster-maintenance/decommission-brokers.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+The decommissioning process is successful only when all partition reallocations have been completed successfully. The controller leader polls for the status of all the partition-level reallocations to ensure that everything completes as expected.
+
+During the decommissioning process, new partitions are not allocated to the broker that is being decommissioned. Once all the reallocations have been completed successfully, the broker is removed from the cluster.
```
micheleRP, line 41:
```suggestion
During the decommissioning process, new partitions are not allocated to the broker that is being decommissioned. After all the reallocations have been completed successfully, the broker is removed from the cluster.
```

docs/manage/kubernetes/decommission-brokers.mdx
```diff
@@ -0,0 +1,158 @@
...
+
+:::info
+When a broker is decommissioned, it cannot rejoin the cluster. If a broker with the same ID tries to rejoin the cluster, it is rejected.
+:::
+
+## What happens when a broker is decommissioned
+
+When a broker is decommissioned, the controller leader creates a reallocation plan for all partition replicas that are allocated to that broker. By default, this reallocation is done in batches of 50 to avoid overwhelming the remaining brokers with Raft recovery. See [`partition_autobalancing_concurrent_moves`](../../../reference/tunable-properties#partition_autobalancing_concurrent_moves).
+
+The reallocation of each partition is translated into a Raft group reconfiguration and executed by the controller leader. The partition leader then handles the reconfiguration for its Raft group. Once the reallocation for a partition is complete, it is recorded in the controller log and the status is updated in the topic tables of each broker.
```
micheleRP, line 43:
```suggestion
The reallocation of each partition is translated into a Raft group reconfiguration and executed by the controller leader. The partition leader then handles the reconfiguration for its Raft group. After the reallocation for a partition is complete, it is recorded in the controller log and the status is updated in the topic tables of each broker.
```

docs/manage/kubernetes/decommission-brokers.mdx
JakeSCahill: I think if we eventually move Kubernetes docs to a separate versioned nav, having this `k` prefix will seem superfluous and confusing. We should probably record this in a separate issue.

docs/manage/decommission-brokers.mdx
JakeSCahill: I think a glossary or the Cluster Balancing page would be better. The goal of users in this doc is to decommission brokers rather than learn the differences between the two leaders. We should have a dedicated conceptual topic for that and link to it.

micheleRP: I'll add in Cluster Balancing page!

docs/manage/kubernetes/decommission-brokers.mdx
micheleRP: We can track with https://github.com/redpanda-data/documentation/issues/1454

PR #1414, opened 2023-03-21
engine: remove beta label

Deflaimun:
Remove beta label as requested by @micheleRP


micheleRP (approved):
Thanks, Paulo, for hiding the label from our dropdown. We'll still use this branch for documenting 23.2 features, but users won't see it until we have a 23.2 download available and we unhide this.

PR #1462, opened 2023-04-03
docs: Document initContainers.setDataOwnership

### LINKED ISSUE #1452: Provide troubleshooting for `fatal error during checker "Data directory is writable"` in Kubernetes
### Describe the Issue

In the Helm chart we [removed the initContainer](https://github.com/redpanda-data/helm-charts/issues/370) because it ran as root and some Kubernetes clusters do not allow Pods to run as root due to security risks. However, a [recent issue](https://github.com/redpanda-data/helm-charts/issues/387) in the Helm chart found that the initContainer is required when running on minikube or using `hostPath` volumes, so we added it back but disabled it by default.

The Pods running Redpanda brokers must have read/write access to their data directories. The initContainer is responsible for setting write permissions on those data directories. This initContainer is disabled by default because most storage drivers call `SetVolumeOwnership` to give permissions to the root of the storage mount. However, `hostPath` drivers do not call `SetVolumeOwnership` because it's expected that `hostPath` is likely to be used for managing the files of the host operating system and you don't just want to change ownership of those files. In this case, users must enable the initContainer with `initContainers.setDataDirOwnership.enabled=true` so that it can set the correct permissions on the data directories. Otherwise, users will see the following error during deployment:

```
Error: fatal error during checker "Data directory is writable" execution: open /var/lib/redpanda/data/test_file: permission denied
```

We must call out that the initContainer runs as root so that users are aware of the security risks. An initContainer running as root can introduce the following security risks:

- Privilege Escalation: If an attacker gains access to the initContainer, they can escalate privileges to gain full control over the system. This could result in unauthorized access to sensitive data, tampering with the system, and causing denial-of-service attacks.
- Container Breakouts: Running an initContainer as root could potentially allow an attacker to escape from the container and access the host operating system. This can happen if there is a vulnerability in the container runtime or if the container is misconfigured.
- Image Tampering: If an attacker gains access to the container image used by the initContainer, they could tamper with the image to add malicious code or backdoors. This could compromise the security of the entire cluster.

JakeSCahill:
Review deadline: April 4

Fixes #1452


joejulian (approved):

docs/manage/kubernetes/shared/_initContainer.mdx
```diff
@@ -0,0 +1,9 @@
+Pods that run Redpanda brokers must have read/write access to their data directories. The initContainer is responsible for setting write permissions on the data directories. By default, `initContainers.setDataDirOwnership` is disabled because most storage drivers call `SetVolumeOwnership` to give Redpanda permissions to the root of the storage mount. However, some storage drivers, such as `hostPath`, do not call `SetVolumeOwnership`. In this case, you must enable the initContainer to set the permissions.
+
+:::info
+To set permissions on the data directories, the initContainer must run as root. An initContainer running as root can introduce the following security risks:
```
Feediver1, line 4:
```suggestion
To set permissions on the data directories, the initContainer must run as root. However, be aware that an initContainer running as root can introduce the following security risks:
```

Issue #1465, opened 2023-04-03
Make external search prioritize the latest version


Deflaimun:
### Describe the Issue

It has been reported multiple times that google search is taking users to different versions than the latest. 

And because of a bug that exists in #1065 it's currently not possible to provide an accurate redirect for all pages.

Deflaimun:
The solution here will be done in 3 steps.

1 - Adding a quick fix solution to the robots.txt file to make google not index versioned pages. This will be removed after 2)

2 - Create canonical links to the latest version page. Remove solution 1 

3 - Migrate those solutions to Antora + solve #1065 through Antora versioning scheme and linking.

Deflaimun:
This issue should only be closed when fully finished.

PR #1466, opened 2023-04-03
engine: disallow crawl through old versions

Deflaimun:
Apply solution 1 from #1465


JakeSCahill:
Google says that `noindex` is the solution. Using `Disallow` will still allow old versions to appear in search results:

3CCB5F5E-2D0B-4076-8014-6D97E9A9D3D8

https://developers.google.com/search/docs/crawling-indexing/robots/intro

Deflaimun:
We have discussed this solution and we're not going for it because we have no control about how often google crawls through our website. 
This quickfix while quick on our side can have long lasting effects for us.

PR #1489, opened 2023-04-10
Docs/#744 doc for reconnections

### LINKED ISSUE #744: Community Feedback: Add doc describing broker configs, re-connections
### Describe the Issue

https://redpandacommunity.slack.com/archives/C01AJDUT88N/p1666678527028509

Comes from RP Community channel question:
Is there a config in redpanda that will handle re-connections to the broker? Like if the broker goes down, it waits for a given amount of time before aborting...

Answer From @jrkinley 
 This is a good question, and the short answer is yes.
The longer answer is if your cluster has 3 brokers then Redpanda can tolerate 1 of those brokers going down. When this happens the partition leaders move to one of the remaining operational brokers and the updated metadata is communicated to the clients, so that they know where to read and write the data. This all happens very quickly and without interruption. If you have a cluster of 5 brokers then it can tolerate 2 failures, etc. As long as the majority of brokers are available, Redpanda is operational.
There are some client configs to be aware of:
message.send.max.retries: how many times a producer will attempt to send a message before marking it as failed
`message.timeout.ms`: in unison with the above, this sets a timeout for delivery so messages are not retried forever. Delivery fails when either the retry count or the message timeout are exceeded
[retry.backoff.ms](http://retry.backoff.ms/): how long to wait in between retries
(edited)
:rocket:
1

sdmg15
(https://redpandacommunity.slack.com/archives/C01AJDUT88N/p1666714694792769?thread_ts=1666678527.028509&cid=C01AJDUT88N)
[@James Kinley](https://redpandacommunity.slack.com/team/U02JT3RGYRF)
 Well, suppose I have only one broker, but it goes down for some reason. My consumer will fail with the error Broker transport Failure But because I don't want my consumer to stop, I need a mechanism to tell him, "Hey even though the only broker we have is down wait for 10min before shutting down if it's sill not up"

James Kinley

Ok, on the consumer side you can set:
reconnect.backoff.ms: amount of time to wait before attempting to reconnect to the broker
reconnect.backoff.max.ms: when this is set, reconnection attempts will backoff exponentially up to this maximum

Joyce Fee
 I looked through our docs and cannot find specific guidance on broker connections--managing and troubleshooting them. Seems like this is a topic we should have. Do you agree? If so, I can take this post and use as the foundation for a doc Issue to get this guidance added.

James Kinley
Hey 
[@Joyce Fee](https://redpandacommunity.slack.com/team/U039T4R4PJ7)
. Yes I agree. It would be great to have a docs page that highlights the important producer and consumer configurations and how they affect operations. I’m happy to contribute too.

sdmg15
The same configs works for both the producer and the consumer? 
[@James Kinley](https://redpandacommunity.slack.com/team/U02JT3RGYRF)

micheleRP:
resolves #744 

preview


jrkinley (changes requested):

docs/manage/cluster-maintenance/configure-availability.mdx
```diff
@@ -11,20 +11,20 @@ import ConfigureAvailabilityIntro from './shared/_intro-configure-availability.m
 
 <ConfigureAvailabilityIntro/>
 
-## Limit Client Connections
+## Limit client connections
 
-A malicious Kafka client application may create many network connections to execute its attacks. A poorly configured application may also create an excessive number of connections. To mitigate the risk of a client creating too many connections and using too many system resources, you can configure a Redpanda cluster to impose limits on the number of created client connections for its Kafka protocol server nodes. 
+A malicious Kafka client application may create many network connections to execute its attacks. A poorly configured application may also create an excessive number of connections. To mitigate the risk of a client creating too many connections and using too many system resources, you can configure a Redpanda cluster to impose limits on the number of created client connections for its Kafka protocol server brokers.
```
JakeSCahill, line 16: > Kafka protocol server brokers

This sounds confusing. Is it referring to the brokers' Kafka listeners?

JakeSCahill (approved):

Feediver1:
@jrkinley I don't think this topic is named accurately. Upon first glance, I would expect some type of high availability config guidance, but that isn't what this is at all. Rather, it is a tuning topic. Question is--what is an accurate description of the thing being tuned? Getting this title right will help customers better understand at a glance exactly what this topic covers. As named, I don't think anyone browsing the navtree would accurately identify what it actually covers. Please advise.

docs/manage/cluster-maintenance/configure-availability.mdx
micheleRP: I'll change this to "...you can configure a Redpanda cluster to impose limits on the number of created client connections."

docs/manage/cluster-maintenance/configure-availability.mdx
```diff
@@ -33,16 +33,35 @@ Redpanda also provides properties to manage the rate of connection creation:
...
-- If the limit is less than two, the node will be blocked from restarting after every single crash, until one of the reset conditions is met.
+- The `crash_loop_limit` property is disabled by default. You must manually enable it by setting it to a non-zero value.
+- If the limit is less than two, the broker is blocked from restarting after every crash, until one of the reset conditions is met.
 :::
+
+## Configure client reconnections
+
+Redpanda’s internal implementation of Raft allows it to tolerate losing a minority of replicas for a given topic partition. For example, if your cluster has three brokers, Redpanda can tolerate one of those brokers going down. If your cluster has five brokers, then it can tolerate two failures, and so on. 
+
+However, if the majority of brokers are unavailable (that is, if Raft cannot form a majority), then producers and consumers will fail. You can configure the Kafka client backoff and retry properties to change the default behavior of the clients to suit your failure requirements.
```
micheleRP, line 54:
```suggestion
You can configure the Kafka client backoff and retry properties to change the default behavior of the clients to suit your failure requirements.
```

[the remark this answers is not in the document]
micheleRP: @mmaslankaprv, @jcsp: Please confirm if it's best to remove the intro bit about Raft (as in suggested change below), and just say, "You can configure the Kafka client backoff and retry properties to change the default behavior of the clients to suit your failure requirements."

micheleRP:
> @jrkinley I don't think this topic is named accurately. Upon first glance, I would expect some type of high availability config guidance, but that isn't what this is at all. Rather, it is a tuning topic. Question is--what is an accurate description of the thing being tuned? Getting this title right will help customers better understand at a glance exactly what this topic covers. As named, I don't think anyone browsing the navtree would accurately identify what it actually covers. Please advise.
> 
> Upon further discussion with @micheleRP , we are really struggling to understand an appropriate context for this topic! I think once it is named correctly, we can then find a better/more suitable context for it. Again--your thoughts @jrkinley .

@jrkinley Instead of Configure Availability as the title, what do you think about Configure Client Connections or Tune Connection Availability? And do you think it's correctly placed under Manage > Cluster Maintenance, or should this page be moved closer to Configure Topics and Produce Data under Develop?

jrkinley (approved):

PR #1511, opened 2023-04-17
docs: Update EKS guide

### LINKED ISSUE #1446: Need a more specific IAM role name used in EKS deployment guide
### Describe the Issue

Following the EKS install guide can potentially cause problems during the helm install step if the IAM role created by the `eksctl create iamserviceaccount` step already exists.   That step will report back with an error about the role already existing, but most users would believe that they could proceed safely with that particular error.

However the role that is created in that step has a trust policy tied to the OIDC provider for the EKS cluster.   If that role already exists (as it does in the Redpanda AWS sandbox-support acct), it will have a trust policy tied to some prior EKS cluster which will cause the helm install to fail with a cryptic timeout error. 

https://docs.redpanda.com/docs/deploy/deployment-option/self-hosted/kubernetes/eks-guide/

The simplest solution is to modify the name of the role to be more user/cluster-specific.   Typically this would be in the form of a command line substitution or an environment variable, but the so-called "in line replacement" would also work and doesn't introduce any os-specific syntax.

From Step 2 in the guide:

```
eksctl create iamserviceaccount \
    `# Do not change the name. It is required by EKS.` \
    --name ebs-csi-controller-sa \
    `# Do not change the namespace. It is required by EKS.` \
    --namespace kube-system \
    --cluster redpanda \
    --attach-policy-arn arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicy \
    --approve \
    --role-only \
    --role-name AmazonEKS_EBS_CSI_DriverRole-<cluster-name>
```

And then in step 3 a corresponding change must be made to account for the new role name.

```
eksctl create addon \
    --name aws-ebs-csi-driver \
    --cluster redpanda \
    --service-account-role-arn arn:aws:iam::${AWS_ACCOUNT_ID}:role/AmazonEKS_EBS_CSI_DriverRole-<cluster-name> \
    --force
```

_It does seem worth noting that an environment variable for the cluster name could be leveraged in several other places in this guide and would ultimately be easier for the end user, although it would make for a more complicated change to the docs, and would probably prompt similar changes to the GKE & EKS guides._

### LINKED ISSUE #1453: Provide command for updating security groups in EKS guide
### Describe the Issue

The guide is very prescriptive in all the steps....except for step 6:

```
Add inbound firewall rules to your EC2 instances so that external traffic can reach the following node ports on all Kubernetes worker nodes in the cluster:

* 31644
* 31092
* 30082
* 30081

For help creating firewall rules, see the [Amazon EC2 documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/authorizing-access-to-an-instance.html).
```

It's fairly trivial to provide the correct build command to the user.

https://docs.redpanda.com/docs/deploy/deployment-option/self-hosted/kubernetes/eks-guide/

`Step 6:  Add inbound firewall rules to your EC2 instances so that external traffic can reach the following node ports on all Kubernetes worker nodes in the cluster` needs to read more like this:

---

Step 6:  Add inbound firewall rules to your EC2 instances so that external traffic can reach the following node ports on all Kubernetes worker nodes in the cluster by first identifying the id of the security group associated with the nodes in your EKS cluster:

`aws eks describe-cluster --name redpanda | jq -r '.cluster.resourcesVpcConfig.clusterSecurityGroupId'`

Then use the security group ID returned in in creating three ingress rules to allow traffic to the worker nodes:

```
aws ec2 authorize-security-group-ingress \
    --group-id <security group id> \
    --ip-permissions "[ \
                        { \
                          \"IpProtocol\": \"tcp\", \
                          \"FromPort\": 30081, \
                          \"ToPort\": 30082, \
                          \"IpRanges\": [{\"CidrIp\": \"0.0.0.0/0\"}] \
                        }, \
                        { \
                          \"IpProtocol\": \"tcp\", \
                          \"FromPort\": 31644, \
                          \"ToPort\": 31644, \
                          \"IpRanges\": [{\"CidrIp\": \"0.0.0.0/0\"}] \
                        }, \
                        { \
                          \"IpProtocol\": \"tcp\", \
                          \"FromPort\": 31092, \
                          \"ToPort\": 31092, \
                          \"IpRanges\": [{\"CidrIp\": \"0.0.0.0/0\"}] \
                        } \
                      ]"

```

_Note this allows traffic from all IPs on ports 30081, 30082, 31644, and 31092.   A better practice would be to restrict inbound traffic to only those IPs from which you expect inbound traffic._

---

### Additional notes

For the IP traffic note, AWS puts it in a warning box with this wording, which we may or may not want to copy directly.

```If you use 0.0.0.0/0, you enable all IPv4 addresses to access your instance using SSH. If you use ::/0, you enable all IPv6 address to access your instance. You should authorize only a specific IP address or range of addresses to access your instance.```

Also, the intent of the `<security group id>` is to use an editable placeholder.   An alternative approach would be to have the describe cluster command export the output to an environment variable, and then leverage that variable in the authorize-security-group-ingress command.   For what it's worth, we mention at the top of the guide that users will need jq installed in order to parse json output and store results in environment variables, but we don't actually use environment variables anywhere.  Using environment variables (here and elsewhere throughout the guide) could make this entire guide completely copy/paste.

### LINKED ISSUE #1464: EKS deployment guide - bad command & example outputs
### Describe the Issue

**Issue 1:** kubectl command for verifying that Redpanda is scheduled on each node is incorrect.  Somewhere along the way a label changed and this command didn't get updated accordingly.

**Issue 2:**  very minor, but the corresponding expected output section isn't fully representative of what the output actually looks like.

**Issue 3:**  the output from `rpk-topic describe` is not exactly what is returned (based on the instructions in the guide).   This is also minor, but the guide appears to show that some other config options have been set that aren't super relevant to the guide itself.

https://docs.redpanda.com/docs/deploy/deployment-option/self-hosted/kubernetes/eks-guide/?operating-system=macos

**Issue 1:  wrong label in the filtering section of the kubectl command under Deploy Redpanda and redpanda console**

https://docs.redpanda.com/docs/deploy/deployment-option/self-hosted/kubernetes/eks-guide/?operating-system=macos#deploy-redpanda-and-redpanda-console

the label used in filter of the given command is `app.kubernetes.io/component=redpanda` but this doesn't return anything because there is no matching label when you do a `kubectl describe pod`.   I believe this should be `app.kubernetes.io/component=redpanda-statefulset` which returns output similar to what is in the Expected Output section.

```
kubectl get pod -n redpanda  \
-o=custom-columns=NODE:.spec.nodeName,NAME:.metadata.name -l \
app.kubernetes.io/component=redpanda-statefulset,job-name!=redpanda-redpanda-configuration,job-name!=redpanda-post-upgrade
```

---

**Issue 2:  Expected Output isn't perfect**

The expected output from the command in Issue 1 above is missing the header row, and the node field could look a little more representative than just "example-worker-0".   This is what the actual output looks like.  Obviously the IP's & AWS region will be unique to each deployment, and I am open to ideas on how to best represent that variability in the documentation.   Although given that it is _example_ output, it's probably ok to leave the domain names exactly as they appear here.

```
NODE                                           NAME
ip-192-168-59-212.us-east-2.compute.internal   redpanda-0
ip-192-168-75-108.us-east-2.compute.internal   redpanda-1
ip-192-168-25-198.us-east-2.compute.internal   redpanda-2
```

---

Issue #3:  rpk-topic describe output isn't 100% correct

This is pretty minor, but as long as we're touching the docs we may as well have this be 100% correct.   If you follow the guide to the letter, the output will look as follows.  The way it currently is looks like somebody made some cluster config changes that aren't accounted for in the guide, and aren't necessary to complete the guide successfully.

```
SUMMARY
=======
NAME        twitch_chat
PARTITIONS  1
REPLICAS    1

CONFIGS
=======
KEY                           VALUE       SOURCE
cleanup.policy                delete      DYNAMIC_TOPIC_CONFIG
compression.type              producer    DEFAULT_CONFIG
max.message.bytes             1048576     DEFAULT_CONFIG
message.timestamp.type        CreateTime  DEFAULT_CONFIG
redpanda.remote.delete        true        DEFAULT_CONFIG
redpanda.remote.read          false       DEFAULT_CONFIG
redpanda.remote.write         false       DEFAULT_CONFIG
retention.bytes               -1          DEFAULT_CONFIG
retention.local.target.bytes  -1          DEFAULT_CONFIG
retention.local.target.ms     86400000    DEFAULT_CONFIG
retention.ms                  604800000   DEFAULT_CONFIG
segment.bytes                 134217728   DEFAULT_CONFIG
segment.ms                    1209600000  DEFAULT_CONFIG
```

JakeSCahill:
Review deadline: April 19

Fixes #1446 , fixes #1453 , fixes #1464 

Preview: https://deploy-preview-1511--redpanda-documentation.netlify.app/docs/deploy/deployment-option/self-hosted/kubernetes/eks-guide/


supahcraig:
Step 3 - "Create the IAM role needed for the Amazon Elastic Block Store (EBS) Cluster Storage Interface (CSI):"
* the substitution variable for cluster name is not present; still hardcoded as `redpanda`

Step 4 - "Get your AWS account ID:"
* the environment variable has a dash, should be `AWS_ACCOUNT_ID`

Step 5 - "Add the EBS CSI add-on:"
* same as above: the environment variable has a dash, should be `AWS_ACCOUNT_ID`

supahcraig:
Last thing, and this is really pedantic but there is a leading space in the substitution variable in step 1, which injects a space into step 3.

supahcraig (approved):
substitution variables seem to have an extra leading space, which can cause a problem if that variable is used elsewhere (as in the case of the cluster name when it is embedded in the IAM role name in step 3)

Feediver1 (approved):
lgtm

PR #1517, opened 2023-04-18
docs:#1469 add guidance on risks with topic recovery

### LINKED ISSUE #1469: cloud_storage: add guidance on risks/caveats around topic recovery
Currently we provide quite brief guidance on how to create a topic that recovers data from object storage:
https://docs.redpanda.com/docs/manage/tiered-storage/#remote-recovery

We need to provide some guidance to users to avoid bad situations if the functionality is misused:
- This is intended for disaster recovery: it is not a general data movement tool
- Doing a recovery with redpanda.remote.write is only safe if there is definitely not some other cluster still writing to the same bucket.
- If a user reads this and goes ahead and tries to recover e.g. every topic from a cluster, the data will not have snapshot-style consistency: they will need to think carefully about how their application would respond to some partitions in some topics being more up to date than others.  If they try to restore the consumer offsets topic, it could have offsets that point ahead of restored data.

Separately, when full metadata recovery is delivered, there would be separate docs content for that.  That will have different caveats, although will still be a disaster recovery tool rather than a "go ahead and try this at will" type thing.

**or**

Maybe we should just pull description of this functionality from the docs and replace it with a note that recoverying data from object storage is possible with Redpanda, and please to contact Redpanda support for help doing that.

micheleRP:
resolves #1469 

This adds a caution about using remote recovery and changes the title of that section from Remote recovery to Disaster recovery.

preview: 
- https://deploy-preview-1517--redpanda-documentation.netlify.app/docs/manage/tiered-storage/#disaster-recovery
- https://deploy-preview-1517--redpanda-documentation.netlify.app/docs/manage/kubernetes/tiered-storage/#disaster-recovery


docs/manage/shared/tiered-storage/_remote-recovery.mdx
```diff
@@ -1,5 +1,9 @@
 When you create a topic, you can use remote recovery to download the topic data from object storage. You can also use remote recovery to restore a topic that was deleted from a cluster.
 
+:::caution
+Recovery with `redpanda.remote.write` is not a general data movement tool. It is only safe when no other clusters are still writing to that bucket or container. Recovery does not provide snapshot-style consistency: some partitions in some topics will be more up to date than others. For example, restoring a consumer offsets topic could have offsets that point ahead of the restored data.
```
Feediver1, line 4:
```suggestion
Recovery with `redpanda.remote.write` is not a recommended general data movement method, and is only safe to use when no other clusters are still writing to the specified bucket or container. Recovery does not provide snapshot-style consistency: some partitions in some topics will be more up to date than others. For example, restoring a consumer offsets topic could have offsets that point ahead of the restored data.
```

Feediver1 (approved):
Left a suggestion

mattschumpert (approved):
'Data Movement' is a weird thing to say.

I would say 'Remote Recovery is not intended to be a fully functional Disaster Recovery solution'

PR #1519, opened 2023-04-19
engine: Improve the editable placeholders script

JakeSCahill:
While implementing this in Antora, I discovered a more efficient way to make the placeholders editable.

Before, the code was running multiple times as part of a useEffect for each CodeBlock. Now, it runs once on page load and the logic to find placeholders is simpler (with a single regex).


Deflaimun:
Good example for testing:
https://deploy-preview-1519--redpanda-documentation.netlify.app/docs/get-started/licenses/?platform=kubernetes

PR #1524, opened 2023-04-19
Document storage_ignore_timestamps_in_future_sec cluster property

### LINKED ISSUE #1522: Document `storage_ignore_timestamps_in_future_sec`
New cluster configuration property: `storage_ignore_timestamps_in_future_sec`

This enables recovering a Redpanda cluster from situations where it is not cleaning up data on disk because the user sent records with timestamps far in the future, so Redpanda thinks the retention time limit hasn't been reached yet.

When this property is set to something (it is null by default), we change our behavior.  For example if the property is set to `60`, then any timestamps more than 60 seconds in the future are ignored by Redpanda when doing retention: it will make its best guess at the age of a segment instead, by either:
- Looking for another record in the same segment that has a valid (within ignore threshold) timestamp
- Looking at the `mtime` of the file itself.

This is a change that is soon to backport into 23.1 and 22.3.

From https://github.com/redpanda-data/redpanda/pull/10028
```
We should probably write down some instructions on how to detect scenarios where bogus timestamps are blocking retention and how to use this new config to resolve the situation.

Probably something along the lines:

- Look for "segment with bogus retention timestamps" logs
- From the log lines above, determine how far into the future those timestamps are
- Set storage_ignore_timestamps_in_future_sec to something that will allow retention to proceed
- Unset storage_ignore_timestamps_in_future_sec if the bogus timestamps was due to a transient client issue (perhaps the test should be extended to do this)
```

shorgi:
Fixes #1522 

Previews:
- [23.1 storage_ignore_timestamps_in_future_sec reference](https://deploy-preview-1524--redpanda-documentation.netlify.app/docs/reference/cluster-properties/#storage_ignore_timestamps_in_future_sec)
- [22.3 storage_ignore_timestamps_in_future_sec reference](https://deploy-preview-1524--redpanda-documentation.netlify.app/docs/22.3/reference/cluster-properties/#storage_ignore_timestamps_in_future_sec)(same content as 23.1)

Blocked from merging until 23.1.8 is released


docs/reference/cluster-properties.mdx
```diff
@@ -1094,6 +1094,29 @@ Requires that an empty file named `.redpanda_data_dir` be present in the [data d
...
+The maximum number of seconds that a record's timestamp can be ahead of a Redpanda broker's clock and still be used when deciding whether to clean up the record for data retention. This property makes possible the timely cleanup of records from clients with clocks that are drastically unsynchronized relative to Redpanda.
+
+When determining whether to clean up a record with timestamp more than `storage_ignore_timestamps_in_future_sec` seconds ahead of the broker, Redpanda ignores the record's timestamp and instead uses a valid timestamp of another record in the same segment, or (if another record's valid timestamp is unavailable) the timestamp of when the segment file was last modified (mtime).
+
+By default, `storage_ignore_timestamps_in_future_sec` is disabled (null).
+
+:::tip
+To figure out whether to set `storage_ignore_timestamps_in_future_sec` for your system:
+1. Look for logs with segments that are unexpectedly large and not being cleaned up.
+1. In the logs, search for records with unsynchronized timestamps that are far into the future.
```
Feediver1, line 1110: Read this trough a couple of times and both times found myself asking, in this context what is the definition of "far into the future"? Is there a future time threshold one should look for--an amount of time/future period that exceeds such a threshold?

docs/reference/cluster-properties.mdx
```diff
@@ -1094,6 +1094,29 @@ Requires that an empty file named `.redpanda_data_dir` be present in the [data d
...
+1. If you find unsynchronized timestamps, determine the number of seconds that the timestamps are ahead of their actual time, and set `storage_ignore_timestamps_in_future_sec` to that value so data retention can proceed.
+1. If the unsynchronized timestamps were the result of transient behavior, you can disable `storage_ignore_timestamps_in_future_sec`.
```
Feediver1, line 1112:
```suggestion
1. If the unsynchronized timestamps are the result of transient behavior, you can disable `storage_ignore_timestamps_in_future_sec`.   
```

Feediver1 (approved):
one comment & suggestion--your call

docs/reference/cluster-properties.mdx
jcsp: It depends how much clock drift a user tolerates on their clients.  Anything more than a second is a bad clock.  But if the user is running something from e.g. a laptop, then the drift might be much more.

If we were to give an  example of "far" then 60 seconds should be uncontroversial

shorgi: @jcsp - I updated the Tip admonition, please review and comment/approve

mattschumpert:
@shorgi FYI this doesn't get into the product until next week, 23.1.8, just making sure we clarify that in docs

shorgi:
> @shorgi FYI this doesn't get into the product until next week, 23.1.8, just making sure we clarify that in docs

@mattschumpert understood, not merging until the release goes live

shorgi:
@mattschumpert with 23.1.8 released ([announcement](https://redpandadata.slack.com/archives/C029XT7FLQ2/p1682665457485899)), I'll be merging this now

Feediver1:
Most PRs require SME approval, but looks like requested changes have been implemented, and this PR has sat for 3 weeks awaiting that approval. @shorgi Please merge. @jcsp if any issues, please open a new ticket. Thx!

PR #1535, opened 2023-04-21
docs:#1358 update doc about Tiered Storage properties

### LINKED ISSUE #786: Open comments for Tiered Storage
### Describe the Issue

The [22.3 Tiered Storage retention controls PR](https://github.com/redpanda-data/documentation/pull/709) has open comments/questions that couldn't be resolved prior to 22.3  This tracks those issues.

[mattschumpert](https://github.com/mattschumpert) 
(https://github.com/redpanda-data/documentation/pull/709#discussion_r1004953657)
@Lazin regarding this sentence: 'Redpanda Tiered Storage works behind the scenes to index where data is offloaded, so it can retrieve the data when you need it.'

What are we really trying to say here? IIUC the 'compressed in memory index' (referred to here https://redpanda.com/blog/tiered-storage-architecture-shadow-indexing-deep-dive). let's us retrieve data 'efficiently' when you need it, because it can access it via a translated file offset and not load the whole file. Is my understanding correct? The value of the indexing is efficient lookups?
=======

"To create a cloud storage bucket and configure Redpanda:"
 [mattschumpert](https://github.com/mattschumpert) 
(https://github.com/redpanda-data/documentation/pull/709#discussion_r1004956528)
@micheleRP no, this isn't necessary with the new retention controls ( since Redpanda will expire the data for your), right @Lazin ?
=======

"Use [IAM roles](../../security/iam-roles) to create a user who can access S3."
[mattschumpert](https://github.com/mattschumpert) 
(https://github.com/redpanda-data/documentation/pull/709#discussion_r1004958692)
@micheleRP no, they shouldn't need this if they are using IAM. It's one or the other (IAM role with permissions, or keys).

if they go the IAM route they also need t get cloud storage into IAM mode as described here: https://deploy-preview-709--redpanda-documentation.netlify.app/docs/security/iam-roles/#configuring-iam-roles).

IAM roles is the happy path and folks should only use keys/secrets as a fallback if they cannot use IAM roles

cc @Lazin
========

"5. Edit the the following required properties with the `rpk cluster config edit` command:"
[mattschumpert](https://github.com/mattschumpert) 
(https://github.com/redpanda-data/documentation/pull/709#discussion_r1004959697)
@micheleRP should show the use of IAM roles rather than keys, or perhaps both approaches (1 example of each).
=======

"The table below gives the outcomes for the possible combinations of cluster-level"
[mattschumpert](https://github.com/mattschumpert) 
(https://github.com/redpanda-data/documentation/pull/709#discussion_r1004969520)
@micheleRP there's gotta be a smoother way to explain the layers here, probably needs rewording.

The 2 points to get across are:

cloud_storage_enabled is like them master switch for the Tiered Storage subsystem, nothing happens without it, all other settings are ignored.
However, this itself does not cause data to be stored in cloud storage until other global or topic settings are set.
We can probably do something succinct.

@[micheleRP](https://github.com/micheleRP) micheleRP [3 days ago](https://github.com/redpanda-data/documentation/pull/709#discussion_r1014504561)
@mattschumpert please see the revised text & let me know if you have other suggestions

@[micheleRP](https://github.com/micheleRP) micheleRP [3 days ago](https://github.com/redpanda-data/documentation/pull/709#discussion_r1014501525)
@mattschumpert Can you pls provide the examples we should use here?
=======

"A topic does not have guaranteed local storage, but you can specify hints to stash a range of data in local storage."
[mattschumpert](https://github.com/mattschumpert) 
(https://github.com/redpanda-data/documentation/pull/709#discussion_r1004975391)
@Lazin technically these are not hints and we do not yet have behavior to avoid disk full situations (with 22.3), right?
cc @jcsp
=======

"In Redpanda Cloud, you must explicitly switch on Tiered Storage for specific topics or globally for the cluster. It is not enabled by default."
[mattschumpert](https://github.com/mattschumpert) 
(https://github.com/redpanda-data/documentation/pull/709#discussion_r1007444427)
is this true @jcsp ?
=======

"Remote write is the process that constantly uploads log segments to cloud storage. The process is created for each partition and runs on the leader node of the partition. It only uploads the segments that contain offsets that are smaller than the last stable offset. This is the largest offset that the client can read. "
[mattschumpert](https://github.com/mattschumpert)
(https://github.com/redpanda-data/documentation/pull/709#discussion_r1007445034)
Shouldn't we say 'latest offset' @mmaslankaprv , not 'largest'?

@[mattschumpert](https://github.com/mattschumpert) mattschumpert [3 days ago](https://github.com/redpanda-data/documentation/pull/709#discussion_r1014430363)
Also, doesn't compaction ONLY happen when the user explicitly decides to use a compacted topic @mmaslankaprv , @Lazin ?
=======

"Normally, when data is evicted locally, it is no longer available. If the consum"
[mattschumpert](https://github.com/mattschumpert)
(https://github.com/redpanda-data/documentation/pull/709#discussion_r1007451711)
@micheleRP , this sentence :
If you use remote read without remote write, there is nothing for Redpanda to read.

was probably written before Read Replicas, which do this kind of thing. @Lazin when you create a read replica topic, does it set redpanda.remote.read under the covers on the RR topic? In this case we might want to give this caveat that only read replica topics should be the ones to have remote read without remote write. please advise.
=======

## Set up
Member
@[Feediver1](https://github.com/Feediver1) Feediver1 [3 days ago](https://github.com/redpanda-data/documentation/pull/709#discussion_r1014519281)
Why isn't this a "Prerequisites" section? Are all these steps required to use Tiered Storage--it's unclear if they are required or not? Or are they the actual set up steps for Tiered Storage? Unclear to me.

### LINKED ISSUE #1243: Tiered Storage clarify tuning smaller log segments
### Describe the Issue

see https://redpandadata.slack.com/archives/C03KAHQ776C/p1676480135535389

Chris Larsen
I'm trying to have NBA validate tiered storage on azure blob and they're not seeing new log segments.. I don't remember which config to tune down to be smaller log segments and have scoured this config list.. nothing is jumping out at me https://docs.redpanda.com/docs/manage/tiered-storage/#tiered-storage-configuration-properties

I feel like there should be a section in the doc that explains what to change to see more log segments coming through and make sure they know it's dependent on a constant stream of data to build up the segment(s) to hit the roll threshold
it's the most common question by far from my experience of people kicking tires

James Kinley
You can change segment.bytes to roll the segments more frequently.
There is also…
Upload interval: (cloud_storage_segment_max_upload_interval_sec) Frequency with which new segments will be generated in tiered storage, even if the local segment has not exceeded the segment size.

segment.bytes is local though right, and then i think that max upload one is what i was looking for
segment.bytes has a cluster level default, but you can override at topic level

NOTE: This may be covered with other Tiered Storage/sizing updates to doc. 

https://docs.redpanda.com/docs/manage/tiered-storage/#tiered-storage-configuration-properties

### LINKED ISSUE #1334: Tiered Storage docs should probably mention retention_local_target_bytes_default and retention_local_target_ms_default
### Describe the Issue
In the section [Manage local capacity for Tiered Storage topics](https://docs.redpanda.com/docs/manage/tiered-storage/#manage-local-capacity-for-tiered-storage-topics) we should probably mention that there are global defaults that can be used to manage all topic local storage settings for every 

https://docs.redpanda.com/docs/reference/cluster-properties/#retention_local_target_ms_default

https://docs.redpanda.com/docs/reference/cluster-properties/#retention_local_target_bytes_default

We seem to hint at this with the following statement:

> Starting in version 22.3, new topics use the default retention values of 1 (local) and 7 days (cloud).

because retention_local_target_ms_default is set to 1 day by default.

We should reword the following

> Starting in version 22.3, new topics use the default retention values of 1 (local) and 7 days (cloud).

to something like:

> Starting in version 22.3, new topics use the default retention values of 1 day for local storage (retention_local_target_ms_default) and 7 days for all storage, including cloud object store (delete_retention_ms).

### LINKED ISSUE #1335: Tiered Storage: retention_local_target_bytes_default/ms_default should also point reader to topic-level settings
### Describe the Issue

The reference entries for both retention_local_target_bytes_default and retention_local_target_ms_default should also direct readers to the topic-level configuration overrides for the cluster-level defaults.

retention_local_target_bytes_default should direct the user to retention.local.target.bytes in the Related Properties section.

retention_local_target_ms_default should direct the user to retention.local.target.ms in the Related Properties section.

Both should say something like the following in the description

This property can be overridden on a per-topic basis by setting retention.local.target.<whatever> within each topic enabled for Tiered Storage.

### LINKED ISSUE #1358: Tiered storage docs refer to properties that don't seem to exist
In <https://docs.redpanda.com/docs/manage/tiered-storage/#tiered-storage-configuration-properties|tiered storage docs > we make the following statement:

&gt; •  If `<http://retention.ms|retention.ms>` or `retention.bytes` is set, it initializes the `remote_retention_` properties in `topic_properties`.
What are the remote_retention_* variables? I can't find a reference to them in either the properties documentation or in rp code itself?

[Slack Message](https://redpandadata.slack.com/archives/C04PD4G731V/p1678123572088319?thread_ts=1678123572.088319&cid=C04PD4G731V)

### LINKED ISSUE #1532: Tiered storage docs have repeated text
### Describe the Issue

In https://docs.redpanda.com/docs/manage/tiered-storage/#enable-tiered-storage, we say:

> Topic-level properties override cluster-level properties. For example, if you have Tiered Storage enabled for the cluster and you create a topic with Tiered Storage turned off, it is also turned off for that topic. When cluster-level properties are changed, the changes apply only to new topics, not existing topics.

Then a bit later pretty much the exact same thing, but with a bit more detail and formatting:

>Topic configuration flags override any cluster-level configuration. For example, for new topics, if [cloud_storage_enable_remote_write](https://docs.redpanda.com/docs/reference/tunable-properties/#cloud_storage_enable_remote_write) is set to true, you can set redpanda.remote.write to false to turn it off for a particular topic. Or, if you have Tiered Storage turned off for the cluster, you can enable it for individual topics with the topic flags. When cluster-level properties are changed, the changes apply only to new topics, not existing topics.

The latter seems the more complete, but either way we can lose one of these

micheleRP:
fixes #1358, fixes #1532, fixes #1334, fixes #1335, fixes #1243, fixes #786 

This fixes several small doc issues with Tiered Storage. It also fixes links. 

preview: 
- https://deploy-preview-1535--redpanda-documentation.netlify.app/docs/manage/tiered-storage/#manage-local-capacity-for-tiered-storage-topics (clarified & added info for `retention_local_target_ms_default` and `retention_local_target_bytes_default`)
- https://deploy-preview-1535--redpanda-documentation.netlify.app/docs/manage/tiered-storage/#enable-tiered-storage (simplified steps, removed redundancy, fixed links)
- https://deploy-preview-1535--redpanda-documentation.netlify.app/docs/reference/cluster-properties/#retention_local_target_bytes_default (can override `retention_local_target_bytes_default` and `retention_local_target_ms_default` with topic properties--topic properties aren't documented live yet, so we can't link to them yet )
- https://deploy-preview-1535--redpanda-documentation.netlify.app/docs/manage/tiered-storage/#remote-write (new TIP at bottom of section)


docs/manage/shared/tiered-storage/_adjacent-segment-merging.mdx
```diff
@@ -3,17 +3,17 @@ import TunableClusterProperty from '../tiered-storage/_tunable-cluster-props.mdx
 
 By default, and as part of housekeeping, Redpanda runs adjacent segment merging on all segments in object storage that are smaller than the threshold. Two properties control the behavior of <ClusterProperty name="cloud_storage_enable_segment_merging" frontmatter={props.frontmatter}/>:
```
micheleRP, line 4: @shorgi can you please add `cloud_storage_enable_segment_merging` to the property reference file, so we can link on this page?

micheleRP: @shorgi also please add `cloud_storage_idle_threshold_rps`!

shorgi: @micheleRP Added props in [854c429](https://github.com/redpanda-data/documentation/pull/1535/commits/854c429e1c7f2df58750773bdae8ba412a8f1a7e)

docs/manage/shared/tiered-storage/_remote-write.mdx
```diff
@@ -24,6 +25,10 @@ rpk topic alter-config <topic_name> --set redpanda.remote.write=true
 
 If remote write is enabled, log segments are not deleted until they’re uploaded to object storage. Because of this, the log segments may exceed the configured retention period until they’re uploaded, so the topic might require more disk space. This prevents data loss if segments cannot be uploaded fast enough or if the retention period is very short.
 
+:::tip
+A constant stream of data is necessary to build up the log segments and roll them into object storage. To see new log segments faster, you can edit the <TunableClusterProperty name="cloud_storage_segment_max_upload_interval_sec" frontmatter={props.frontmatter}/> property. This sets the frequency with which new segments are generated in Tiered Storage, even if the local segment has not exceeded the `segment.bytes` size.
```
mattschumpert, line 29: @micheleRP 'exceeded the segment.bytes size or the segment.ms age'.

With 23.1 we now support segment.bytes so this is another way segments get automatically rolled.

mattschumpert (approved):
Approved with 1 minor comment on the TIP

docs/manage/shared/tiered-storage/_enable.mdx
```diff
@@ -3,7 +3,11 @@ import TunableClusterProperty from '../tiered-storage/_tunable-cluster-props.mdx
...
+For either a cluster or a topic, the cluster-level property `cloud_storage_enabled` must be set to `true`. To disable Tiered Storage at the cluster level and enable it on specific topics, this property must still be set to `true`.
+- If this property is set to `false`, nothing is added to object storage, regardless of whether or not the other Tiered Storage properties are enabled.
+- If this property is set to `true` and the other Tiered Storage properties are disabled, then the Tiered Storage subsystem is initialized, but it is not used until you enable Tiered Storage for a topic or at the cluster level.
+
+Topic-level properties override cluster-level properties. For example, for new topics, if `cloud_storage_enable_remote_write` is set to `true`, you can set `redpanda.remote.write` to `false` to turn it off for a particular topic. Or, if you have Tiered Storage turned off for the cluster, you can enable it for individual topics with the topic properties. When cluster-level properties are changed, the changes apply only to new topics, not existing topics.
```
hcoyote, line 10: This is a little confusing.

I think this means that you must have cloud_storage_enabled=true _and_ either 

- cloud_storage_enabled_remote_write=true to affect all topics globally
- redpanda.remote.write=true set within each topic that needs it specifically enabled.

So something like(?):

> To enable Tiered Storage support, cloud_storage_enabled must be set to true. Once this is done, you can either configure all topics globally to write to Tiered Storage or configure only specific topics. To configure this globally, set cloud_storage_enable_remote_write to true. To configure only specific topics, set redpanda.remote.write=true within each topic's configuration. 

I think cloud_storage_enabled is the feature flag that tells Redpanda it is now capable of using tiered storage, whereas the other two are the topic settings that instruct it to actually offload (whether that's globally for all topics or for individual topics)

docs/manage/shared/tiered-storage/_set-retention-limits.mdx
```diff
@@ -14,21 +14,21 @@ During upgrade, Redpanda preserves retention settings for existing topics.
...
+  - Starting in version 22.3, new topics use the default retention values of one day for local storage (`retention_local_target_ms_default`) and seven days for all storage, including object storage (`delete_retention_ms`).
   - Upgraded topics retain their historical defaults of infinite retention.
 
 After topic configuration, if Tiered Storage was disabled and must be enabled, or was enabled and must be disabled, Redpanda uses the local retention properties from the `topic_properties`.
 
 #### Compacted topics in Tiered Storage
 
-To save space, Redpanda compacts segments after they have been uploaded to object storage. Redpanda initially uploads all (non-compacted) segments. It then reuploads the segments with compaction applied.
\ No newline at end of file
+For compacted topics, Redpanda compacts segments after they have been uploaded to object storage. Redpanda initially uploads all (non-compacted) segments. It then reuploads the segments with compaction applied.
```
hcoyote, line 34: does this clean up uncompacted segments from Tiered Storage, too? As written, a reader could walk away with the idea that they may have both uncompacted and compacted segments in tiered storage.

hcoyote (changes requested):

docs/manage/shared/tiered-storage/_enable.mdx
micheleRP: Thanks @hcoyote! I simplified the steps altogether. Please review: https://deploy-preview-1535--redpanda-documentation.netlify.app/docs/manage/tiered-storage/#enable-tiered-storage

docs/manage/shared/tiered-storage/_enable.mdx
```diff
@@ -1,38 +1,54 @@
...
+
+```bash
+rpk topic create <topic_name> -c redpanda.remote.read=true -c redpanda.remote.write=true
+```
+
+To enable Tiered Storage on an existing topic, run both of these commands:
+
+```bash
+rpk topic alter-config <topic_name> --set redpanda.remote.read=true &&
+rpk topic alter-config <topic_name> --set redpanda.remote.write=true
```
VladLazar, line 35: nit: Pretty sure this can be just one command:
```
rpk topic alter-config <topic_name> --set redpanda.remote.read=true --set redpanda.remote.write=true
```

docs/manage/shared/tiered-storage/_enable.mdx
```diff
@@ -1,38 +1,54 @@
...
-When you enable Tiered Storage for a cluster, it is enabled for all existing topics in the cluster.
+To enable Tiered Storage for a cluster (in addition to setting `cloud_storage_enabled` to `true`), set the following cluster-level properties to `true`:
 
-To enable Tiered Storage for a cluster, set the following cluster-level properties to `true`:
-
-- <ClusterProperty name="cloud_storage_enabled" frontmatter={props.frontmatter}/>
 - <TunableClusterProperty name="cloud_storage_enable_remote_write" frontmatter={props.frontmatter}/>
 - <TunableClusterProperty name="cloud_storage_enable_remote_read" frontmatter={props.frontmatter}/>
 
-#### Enable Tiered Storage for topics
```
VladLazar, line 18: I feel like this section would benefit from specifying that `cloud_storage_enable_remote_write` and `cloud_storage_enable_remote_read` are essentially creation-time defaults for the `redpanda.remote.write` and `redpanda.remote.read` topic props.

docs/manage/shared/tiered-storage/_remote-write.mdx
VladLazar: > does this imply that we're uploading the active segment before it's been closed and rolled? Or that TS is force rolling the segment when cloud_storage_segment_max_upload_interval_sec is also reached?

The tiered storage write path can upload active segments. A segment roll is not forced to achieve this. This will happen if `cloud_storage_segment_max_upload_interval_sec` is set, the specified interval has elapsed and all other closed segments have been uploaded.

docs/manage/shared/tiered-storage/_set-retention-limits.mdx
```diff
@@ -14,21 +14,21 @@ During upgrade, Redpanda preserves retention settings for existing topics.
 
 #### Manage local capacity for Tiered Storage topics
 
-You can use topic-level properties to control retention of topic data in local storage. Settings can depend on the size of your drive, how many partitions you have, and how much data you keep for each partition.
+You can use properties to control retention of topic data in local storage. With Tiered Storage enabled, data in local storage expires after the topic-level properties `retention.local.target.ms` or `retention.local.target.bytes`. (These settings are equivalent to `retention.ms` and `retention.bytes` without Tiered Storage.)
```
VladLazar, line 17: nit: I don't know if `expires` is the right word here. I'd say `becomes eligible for deletion` instead.

docs/manage/shared/tiered-storage/_set-retention-limits.mdx
```diff
@@ -14,21 +14,21 @@ During upgrade, Redpanda preserves retention settings for existing topics.
...
 
-With Tiered Storage enabled, data in local storage expires after `retention.local.target.ms` or `retention.local.target.bytes`. These settings are equivalent to `retention.ms` and `retention.bytes` without Tiered Storage.
+You can also use the cluster-level properties `retention_local_target_ms_default` and `retention_local_target_bytes_default`. Settings can depend on the size of your drive, how many partitions you have, and how much data you keep for each partition.
 
 When set, Redpanda keeps actively-used and sequential (next segment) data in local cache and targets to maintain this age of data in local storage. It purges data based on actual available local volume space, without forcing disk full situations when there is data skew. 
 
 At topic creation with Tiered Storage enabled:
-- If `retention.ms` or `retention.bytes` is set, it initializes the `remote_retention_` properties in `topic_properties`.
+- If `retention.ms` or `retention.bytes` is set, it initializes the `retention.local.target.*` properties. 
 - If `retention.local.target.ms` or `retention.local.target.bytes` is set, it is normalized and used to initialize the retention fields in `topic_properties`.
```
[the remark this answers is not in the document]
VladLazar: I think it refers to the fact that when local retention is performed, we will use `min(retention.bytes, retention.local.target.bytes)` and `max(retention.ms, retention.local.target.ms)`

docs/manage/shared/tiered-storage/_set-retention-limits.mdx
VladLazar: > does this clean up uncompacted segments from Tiered Storage, too?

Yes. After Redpanda performs a compacted segment re-upload, the replaced segments will eventually be deleted. This happens when cloud retention is applied, so it's not immediate.

> As written, a reader could walk away with the idea that they may have both uncompacted and compacted segments in tiered storage.

That's pretty much always the case. Compacted re-upload is done opportunistically (when there's no non-compacted uploads to perform; i.e. the entire local log has been uploaded). Hence, it's rare for all segments in the cloud to be compacted. This is perfectly fine though. The tiered storage read path knows how to handle this.

docs/manage/shared/tiered-storage/_remote-write.mdx
micheleRP: I've updated the [tip](https://deploy-preview-1535--redpanda-documentation.netlify.app/docs/manage/tiered-storage/#remote-write). Is this correct?
_A constant stream of data is necessary to build up the log segments and roll them into object storage. To see new log segments faster, you can edit the `segment.bytes` topic-level property. Or, you can edit the `cloud_storage_segment_max_upload_interval_sec` cluster-level property, which sets the frequency with which new segments are generated in Tiered Storage, even if the local segment has not exceeded the `segment.bytes` size or the `segment.ms` age._

docs/manage/shared/tiered-storage/_set-retention-limits.mdx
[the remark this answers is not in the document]
micheleRP: Please review this edit: 
_- If `retention.ms` or `retention.bytes` is set, Redpanda initializes the `retention.local.target.*` properties. 
- If `retention.local.target.ms` or `retention.local.target.bytes` is set, Redpanda initializes the `min(retention.bytes, retention.local.target.bytes)` and `max(retention.ms, retention.local.target.ms)` properties._

docs/manage/shared/tiered-storage/_set-retention-limits.mdx
hcoyote: > Yes. After Redpanda performs a compacted segment re-upload, the replaced segments will eventually be deleted. This happens when cloud retention is applied, so it's not immediate.

I'd call this out. This means we'd have a higher storage cost for long retention periods where compaction is also used.

hcoyote (approved):

docs/manage/shared/tiered-storage/_set-retention-limits.mdx
micheleRP: Revised to: 
_For compacted topics, Redpanda compacts segments after they have been uploaded to object storage. Redpanda initially uploads all uncompacted segments. It then reuploads the segments with compaction applied. It's likely that some segments in object storage are not compacted, but the Tiered Storage read path can manage this._

docs/manage/shared/tiered-storage/_set-retention-limits.mdx
```diff
@@ -4,7 +4,7 @@ Redpanda supports retention limits and compaction for topics using Tiered Storag
 
 Starting in Redpanda version 22.3, object storage is the default storage tier for all streaming data, and retention properties work the same for Tiered Storage topics and local storage topics. Data is retained in object storage until it reaches the configured time or size limit. 
 
-Data expires from object storage following `retention.ms` or `retention.bytes`. For example, if `retention.bytes` is set to 10 GiB, then every partition in the topic has a limit of 10 GiB storage in object storage. When `retention.bytes` is exceeded by data in S3, the data in S3 is trimmed. If neither `retention.ms` nor `retention.bytes` are specified, then cluster level defaults are used.
+Data becomes eligible for deletion from object storage following `retention.ms` or `retention.bytes`. For example, if `retention.bytes` is set to 10 GiB, then every partition in the topic has a limit of 10 GiB storage in object storage. When `retention.bytes` is exceeded by data in S3, the data in S3 is trimmed. If neither `retention.ms` nor `retention.bytes` are specified, then cluster-level defaults are used.
```
VladLazar, line 7: nit: If `cleanup.policy == compact` for the topic, then we won't actually delete anything. Maybe this wording can still be improved somehow.

VladLazar (approved):

docs/manage/kubernetes/tiered-storage.mdx
```diff
@@ -67,7 +67,12 @@ kubectl exec redpanda-0 -c redpanda -n redpanda -- rpk cluster license info
 
 ### Configure object storage
 
-You must configure access to object storage on one of the following supported platforms.
+Redpanda natively supports Tiered Storage with Amazon Simple Storage Service (S3), Google Cloud Storage (GCS, which uses the Google Cloud Platform S3 API), and Microsoft Azure Blob Storage (ABS). 
+
+:::note
```
Feediver1, line 72: This seems important...(rather than a note).

micheleRP: yes, good idea. I made it an important admonition.

docs/manage/shared/tiered-storage/_adjacent-segment-merging.mdx
```diff
@@ -1,19 +1,19 @@
 import ClusterProperty from '../tiered-storage/_cluster-props.mdx'
 import TunableClusterProperty from '../tiered-storage/_tunable-cluster-props.mdx'
 
-By default, and as part of housekeeping, Redpanda runs adjacent segment merging on all segments in object storage that are smaller than the threshold. Two properties control the behavior of <ClusterProperty name="cloud_storage_enable_segment_merging" frontmatter={props.frontmatter}/>:
+By default, and as part of housekeeping, Redpanda runs adjacent segment merging on all segments in object storage that are smaller than the threshold. Two properties control the behavior of <TunableClusterProperty name="cloud_storage_enable_segment_merging" frontmatter={props.frontmatter}/>:
```
Feediver1, line 4: What is the value-add for readers when you say "as part of housekeeping"? This is run regularly? If so, at what interval?

docs/manage/shared/tiered-storage/_adjacent-segment-merging.mdx
```diff
@@ -1,19 +1,19 @@
...
+- <TunableClusterProperty name="cloud_storage_segment_size_target" frontmatter={props.frontmatter}/>: This is the desired segment size in object storage. The default segment size is the local segment size for the topic, which is controlled by the <TunableClusterProperty name="log_segment_size" frontmatter={props.frontmatter}/> configuration property and the <code>segment.bytes</code> topic property. You can set the <code>cloud_storage_segment_size_target</code> property to a value larger than this default segment size, but because that triggers a lot of segment reuploads, it's not recommended.
```
Feediver1, line 6:
```suggestion
- <TunableClusterProperty name="cloud_storage_segment_size_target" frontmatter={props.frontmatter}/>: The desired segment size in object storage. The default segment size is the local segment size for the topic, which is controlled by the <TunableClusterProperty name="log_segment_size" frontmatter={props.frontmatter}/> configuration property and the <code>segment.bytes</code> topic property. You can set the <code>cloud_storage_segment_size_target</code> property to a value larger than this default segment size, but because this triggers a lot of segment re-uploads, it's not recommended. 
```

docs/manage/shared/tiered-storage/_adjacent-segment-merging.mdx
```diff
@@ -1,19 +1,19 @@
...
+- <TunableClusterProperty name="cloud_storage_segment_size_target" frontmatter={props.frontmatter}/>: This is the desired segment size in object storage. The default segment size is the local segment size for the topic, which is controlled by the <TunableClusterProperty name="log_segment_size" frontmatter={props.frontmatter}/> configuration property and the <code>segment.bytes</code> topic property. You can set the <code>cloud_storage_segment_size_target</code> property to a value larger than this default segment size, but because that triggers a lot of segment reuploads, it's not recommended. 
   
-- <ClusterProperty name="cloud_storage_segment_size_min" frontmatter={props.frontmatter}/>: This is the smallest segment size that Redpanda keeps in object storage. The default size is 50% of <TunableClusterProperty name="log_segment_size" frontmatter={props.frontmatter}/>.
+- <TunableClusterProperty name="cloud_storage_segment_size_min" frontmatter={props.frontmatter}/>: This is the smallest segment size that Redpanda keeps in object storage. The default size is 50% of <TunableClusterProperty name="log_segment_size" frontmatter={props.frontmatter}/>.
```
Feediver1, line 8:
```suggestion
- <TunableClusterProperty name="cloud_storage_segment_size_min" frontmatter={props.frontmatter}/>: The smallest segment size that Redpanda keeps in object storage. The default is 50% of <TunableClusterProperty name="log_segment_size" frontmatter={props.frontmatter}/>.
```

docs/manage/shared/tiered-storage/_adjacent-segment-merging.mdx
```diff
@@ -1,19 +1,19 @@
...
 
 If the adjacent segment merging job finds a run of small segments, it can perform one of the following operations:
 
-- Merge and reupload a segment with a size up to <ClusterProperty name="cloud_storage_segment_size_target" frontmatter={props.frontmatter}/>.
-- Merge and reupload a segment with a size less than <ClusterProperty name="cloud_storage_segment_size_min" frontmatter={props.frontmatter}/> if there are no other options (the run of small segments is followed by the large segment).
+- Merge and reupload a segment with a size up to `cloud_storage_segment_size_target`.
+- Merge and reupload a segment with a size less than `cloud_storage_segment_size_min` if there are no other options (the run of small segments is followed by the large segment).
```
Feediver1, line 13:
```suggestion
- Merge and re-upload a segment with a size up to `cloud_storage_segment_size_target`.
- Merge and re-upload a segment with a size less than `cloud_storage_segment_size_min` if there are no other options (the run of small segments is followed by the large segment).
```

docs/manage/shared/tiered-storage/_adjacent-segment-merging.mdx
```diff
@@ -1,19 +1,19 @@
...
+Suppose the <TunableClusterProperty name="cloud_storage_segment_max_upload_interval_sec" frontmatter={props.frontmatter}/> property is set and the partition contains a large number of small segments. For example, if `cloud_storage_segment_max_upload_interval_ms` is set to 10 minutes and the produce rate is 1 MiB per minute, then Redpanda uploads a new 10 MiB segment every 10 minutes. If adjacent segment merging is enabled and `cloud_storage_segment_size_target` is set to 500 MiB, then every 50 segments are reuploaded as one large 500 MiB segment. This doubles the amount of data that Redpanda uploads to object storage, but it also reduces the memory footprint of the partition, which results in better scalability. This is because 98% less memory is needed to keep information about the uploaded segment.
```
Feediver1, line 16:
```suggestion
Suppose the <TunableClusterProperty name="cloud_storage_segment_max_upload_interval_sec" frontmatter={props.frontmatter}/> property is set and the partition contains a large number of small segments. For example, if `cloud_storage_segment_max_upload_interval_ms` is set to 10 minutes and the produce rate is 1 MiB per minute, then Redpanda uploads a new 10 MiB segment every 10 minutes. If adjacent segment merging is enabled and `cloud_storage_segment_size_target` is set to 500 MiB, then every 50 segments are re-uploaded as one large 500 MiB segment. This doubles the amount of data that Redpanda uploads to object storage, but it also reduces the memory footprint of the partition, which results in better scalability because 98% less memory is needed to keep information about the uploaded segment.
```

docs/manage/shared/tiered-storage/_configuration-properties.mdx
```diff
@@ -50,7 +50,7 @@ Under normal circumstances, you should not need to configure the following tunab
...
 | <TunableClusterProperty name="cloud_storage_upload_ctrl_max_shares" frontmatter={props.frontmatter}/> | The maximum number of I/O and CPU shares that the remote write process can use. Default is 1000. |
-| <TunableClusterProperty name="cloud_storage_disable_tls" frontmatter={props.frontmatter}/> | Disables TLS encryption. You can set this to true if TLS termination is done by the proxy, such as HAProxy. Default is false.  |
-| <TunableClusterProperty name="cloud_storage_api_endpoint_port" frontmatter={props.frontmatter}/> | Overrides the default API endpoint port. Default is 443. |
-| <TunableClusterProperty name="cloud_storage_trust_file" frontmatter={props.frontmatter}/> | The public certificate used to validate the TLS connection to object storage. If this is empty, Redpanda uses your operating system's CA cert pool. |
+| <ClusterProperty name="cloud_storage_disable_tls" frontmatter={props.frontmatter}/> | Disables TLS encryption. You can set this to true if TLS termination is done by the proxy, such as HAProxy. Default is false.  |
```
Feediver1, line 53:
```suggestion
| <ClusterProperty name="cloud_storage_disable_tls" frontmatter={props.frontmatter}/> | Disables TLS encryption. Set to true if TLS termination is done by the proxy, such as HAProxy. Default is false.  |
```

docs/manage/shared/tiered-storage/_enable.mdx
```diff
@@ -1,38 +1,57 @@
 import ClusterProperty from '../tiered-storage/_cluster-props.mdx'
 import TunableClusterProperty from '../tiered-storage/_tunable-cluster-props.mdx'
 
-You can enable Tiered Storage for a cluster or a topic.
+1. To enable Tiered Storage, set <ClusterProperty name="cloud_storage_enabled" frontmatter={props.frontmatter}/> to `true`. 
 
-Topic-level properties override cluster-level properties. For example, if you have Tiered Storage enabled for the cluster and you create a topic with Tiered Storage turned off, it is also turned off for that topic. When cluster-level properties are changed, the changes apply only to new topics, not existing topics.
+2. Then, configure topics for Tiered Storage. You can configure either all topics in a cluster or just specific topics.
```
Feediver1, line 6:
```suggestion
2. Configure topics for Tiered Storage. You can configure either all topics in a cluster or just specific topics. 
```

docs/manage/shared/tiered-storage/_enable.mdx
```diff
@@ -1,38 +1,57 @@
...
+| `redpanda.remote.recovery` | Recovers or reproduces a topic from object storage. Use this property during topic creation. It does not apply to existing topics.  |
 | `redpanda.remote.delete` | When set to `true`, deleting a topic also deletes its objects in object storage. Both `redpanda.remote.write` and `redpanda.remote.read` must be enabled, and the topic must not be a Remote Read Replica topic.<br/><br/> When set to `false`, deleting a topic does not delete its objects in object storage.  <br/><br/>Default is `true` for new topics.|
 
-Topic configuration flags override any cluster-level configuration. For example, for new topics, if <TunableClusterProperty name="cloud_storage_enable_remote_write" frontmatter={props.frontmatter}/> is set to `true`, you can set `redpanda.remote.write` to `false` to turn it off for a particular topic. Or, if you have Tiered Storage turned off for the cluster, you can enable it for individual topics with the topic flags. When cluster-level properties are changed, the changes apply only to new topics, not existing topics.
+**Table 1: Remote write configuration options**
```
Feediver1, line 52: I've not seen tables numbered in this way before...what is the value-add for readers? Are the only numbered tables those with config options? Do you think this will be apparent to readers?

docs/manage/shared/tiered-storage/_intro.mdx
```diff
@@ -5,9 +5,7 @@ import EnterpriseLicenseNote from '@site/docs/shared/_enterprise-license.mdx'
 <EnterpriseLicenseNote frontmatter={props.frontmatter}/>
 :::
 
-Tiered Storage helps to lower storage costs by offloading log segments to object storage. You can specify the amount of storage you want to retain in local storage. You don't need to specify which log segments you want to move, because Redpanda moves them automatically based on cluster-level configuration properties. Tiered Storage indexes where data is offloaded, so it can retrieve the data when you need it.
-
-Redpanda natively supports Tiered Storage with Amazon Simple Storage Service (S3), Google Cloud Storage (GCS, which uses the Google Cloud Platform S3 API), and Microsoft Azure Blob Storage (ABS). Migrating topics from one object storage provider to another is not supported. 
+Tiered Storage helps to lower storage costs by offloading log segments to object storage. You can specify the amount of storage you want to retain in local storage. You don't need to specify which log segments you want to move, because Redpanda moves them automatically based on cluster-level configuration properties. Tiered Storage indexes where data is offloaded, so it can lookup the data when you need it.
```
Feediver1, line 8:
```suggestion
Tiered Storage helps to lower storage costs by offloading log segments to object storage. You can specify the amount of storage you want to retain in local storage. You don't need to specify which log segments you want to move, because Redpanda moves them automatically based on cluster-level configuration properties. Tiered Storage indexes where data is offloaded, so it can look up the data when you need it.
```

docs/manage/shared/tiered-storage/_set-retention-limits.mdx
```diff
@@ -4,31 +4,31 @@ Redpanda supports retention limits and compaction for topics using Tiered Storag
...
 :::
 
+Data becomes eligible for deletion from object storage following `retention.ms` or `retention.bytes`. For example, if `retention.bytes` is set to 10 GiB, then every partition in the topic has a limit of 10 GiB in object storage. When `retention.bytes` is exceeded by data in object storage, the data in object storage is trimmed. If neither `retention.ms` nor `retention.bytes` is specified, then cluster-level defaults are used. For more information, see <Link to={props.frontmatter.linkRoot + 'manage/cluster-maintenance/disk-utilization/#configure-message-retention'}>Configure message retention</Link>.
+
+#### Compacted topics in Tiered Storage
+
+When you set <Link to={props.frontmatter.linkRoot + 'develop/config-topics/#change-the-cleanup-policy'}>`cleanup.policy=compact`</Link> for a topic, nothing gets deleted from object storage. 
+
+For compacted topics, Redpanda compacts segments after they have been uploaded to object storage. Redpanda initially uploads all uncompacted segments. It then reuploads the segments with compaction applied. It's likely that some segments in object storage are not compacted, but the Tiered Storage read path can manage this.
```
Feediver1, line 17:
```suggestion
For compacted topics, Redpanda compacts segments after they have been uploaded to object storage. Redpanda initially uploads all uncompacted segments. It then re-uploads the segments with compaction applied. It's likely that some segments in object storage are not compacted, but the Tiered Storage read path can manage this. 
```

Feediver1 (approved):
See suggestions--your call

docs/manage/shared/tiered-storage/_enable.mdx
micheleRP: Good point. I removed the table titles and updated column headings.

docs/manage/shared/tiered-storage/_adjacent-segment-merging.mdx
micheleRP: clarified this

PR #1553, opened 2023-05-01
engine: fix possible form escape.

Deflaimun:


Feediver1 (approved):
So this will keep people from submitting feedback without a URL?

Feediver1:
tested?

Deflaimun:
Not going forward with this change.

PR #1592, opened 2023-05-17
Docs: #1450 changes to local retention behavior

### LINKED ISSUE #1450: Document changes to local retention behaviour
### Describe the Issue

Prior to v23.2, time-based retention on the local log was not applied until size-based retention was no longer applicable. However, the semantics of kafka is "let both policies be active simultaneously" (Kafka code comment [here](https://github.com/apache/kafka/blob/44e613c4cd1a2b7303da7dd30d47bbe87090131f/core/src/main/scala/kafka/log/UnifiedLog.scala#L1358-L1363)). This means that size-based retention will override time-based retention. For example, if in order to satisfy the size based retention policy we must remove 1 segment and in order to satisfy the time based retention we must remove 3 segments, only 1 segment will be removed (according to the size based policy).

I v23.2, both retention policies are applied simultaneously. If we take the previous example again, we'd now remove 3 segments (according to the size based policy).

This change in behaviour needs to be called out somewhere in the docs.

### If new feature, in which release is it included or expected? 
v23.2

### Additional notes
https://github.com/redpanda-data/redpanda/pull/9487

micheleRP:
resolves #1450 

This adds the following note about retention policies:

_Both size-based and time-based retention policies are applied simultaneously, so it's possible for your size-based property to override your time-based property, or vice versa. For example, if your size-based property requires removing one segment, and your time-based property requires removing three segments, then three segments are removed. Size-based properties reclaim disk space as close as possible to the maximum size, without exceeding the limit._

It also changes "upload controller" in Tiered Storage to "upload backlog controller" to differentiate it from the startup controller we'll document with upcoming cluster metadata recovery feature. 

preview: _This may default to 23.1. Make sure to select 23.2 Beta in the dropdown!_
- https://deploy-preview-1592--redpanda-documentation.netlify.app/docs/manage/cluster-maintenance/disk-utilization/#configure-message-retention
- https://deploy-preview-1592--redpanda-documentation.netlify.app/docs/manage/tiered-storage/#set-retention-limits


Feediver1 (approved):
lgtm

mattschumpert (approved):

PR #1637, opened 2023-05-29
Connectors Docs: Added CloudEvents format options to Debezium managed connectors

bochenekmartin:
Debezium MySQL and PostgreSQL managed connectors got `CloudEvents` message key and value format.
For this reason, the documentation for both connectors needs to be extended. 
The CloudEvents format feature is already merged, but not yet available in any environment, so no pressure with merging the documentation update.

Screenshot 2023-05-29 at 11 43 31


vladoschreiner:
👍

Feediver1 (approved):
Thanks for the update @bochenekmartin

PR #1646, opened 2023-05-31
Docs/#1552 document controller snapshots

micheleRP:
resolves #1552 

This adds a new, small conceptual section called "Controller partition and snapshots" to the How Redpanda Works page. 

preview: https://deploy-preview-1646--redpanda-documentation.netlify.app/docs/beta/get-started/architecture/#controller-partition-and-snapshots

PR #1658, opened 2023-06-01
docs:#1552 doc controller snapshots

### LINKED ISSUE #1552: Document Controller Snapshots (23.2)
https://github.com/redpanda-data/core-internal/issues/132

Controller snapshots are somewhat of an implementation detail, but they'll show up for cluster admins and in support incidents as much faster cluster restart / start times, drastically reduce time to add a new node, etc. . This is good for admins to understand (as compared to the previous 'always load the entire controller log on startup' behavior of 23.1 and before)

Snapshots will eventually be used in our metadata backup/DR efforts , but this doc ticket is purely about highlighting how snapshotting works and how redpanda's controller log is managed in general (we can now load the snapshot and replay only incremental updates since the last snapshot to have new and restarted nodes become operational sooner

micheleRP:
resolves #1552 

This was originally in #1646, but since Beta docs already published, I've closed that PR and moved the changes into dev under `versioned_docs/version-23.2`.

This adds a new, small conceptual section called "Controller partition and snapshots" to the How Redpanda Works page.

Preview: https://deploy-preview-1658--redpanda-documentation.netlify.app/docs/beta/get-started/architecture/#controller-partition-and-snapshots


versioned_docs/version-23.2/get-started/architecture.mdx
```diff
@@ -48,6 +48,12 @@ As long as there is a timing inequality between heartbeat time, election timeout
...
 
+## Controller partition and snapshots
+
+Redpanda stores metadata update commands (such as creating and deleting topics or users) in a system partition called the controller partition. Controller snapshots save the current cluster metadata state to disk, so restoring from this snapshot only requires time proportional to the number of metadata objects. A new snapshot is created after each controller command is added. 
+
+Controller snapshots are enabled by default in new clusters and are controlled by the `controller_snapshots` flag. Each broker has a snapshot file stored in the controller log directory: `/var/lib/redpanda/data/redpanda/controller/0_0/snapshot`. The controller partition is replicated by a Raft group that includes all cluster brokers, and the controller snapshot is the Raft snapshot for this group. Snapshots are hydrated when a broker joins the cluster or restarts.
```
ztlpn, line 55: > controlled by the `controller_snapshots` flag

We should mention that this is a feature flag (as opposed to a configuration flag). Unfortunately the feature flag system is not documented...

versioned_docs/version-23.2/get-started/architecture.mdx
```diff
@@ -48,6 +48,12 @@ As long as there is a timing inequality between heartbeat time, election timeout
 
 If a follower triggers an election, but the incumbent leader subsequently springs back to life and starts sending data again, then it’s too late. As part of the election process, the follower (now an election candidate) incremented the term and rejects requests from the previous term, essentially forcing a leadership change. If a cluster is experiencing wider network infrastructure problems that result in latencies above the heartbeat timeout, then back-to-back election rounds can be triggered. During this period, unstable Raft groups may not be able to form a quorum. This results in partitions rejecting writes, but no data loss can occur. Redpanda has a Raft-priority implementation that allows the system to settle quickly after network outages.
 
+## Controller partition and snapshots
+
+Redpanda stores metadata update commands (such as creating and deleting topics or users) in a system partition called the controller partition. Controller snapshots save the current cluster metadata state to disk, so restoring from this snapshot only requires time proportional to the number of metadata objects. A new snapshot is created after each controller command is added.
```
ztlpn, line 53: > A new snapshot is created after each controller command is added.

Stated like this it looks dangerous - a reader might think that Redpanda spams with snapshot on each command. So I am in favor of mentioning debouncing here (maybe not using this exact word, but mentioning that we wait for some period of time in case there are several rapid updates).

versioned_docs/version-23.2/get-started/architecture.mdx
micheleRP: Thanks @ztlpn, I'll add that. And can you provide a Redpanda definition of a feature flag vs config flag? This "How Redpanda Works" page would be a good place to include it.

versioned_docs/version-23.2/get-started/architecture.mdx
micheleRP: Thanks @ztlpn. Please confirm if this update is correct: 
_... Controller snapshots save the current cluster metadata state to disk, so startup is fast. Restoring from this snapshot only requires the time proportional to the number of metadata objects. A new snapshot is created after each controller command is added, or, with rapid updates, after a set period of time (default is 60 seconds)._

versioned_docs/version-23.2/get-started/architecture.mdx
ztlpn: here are some internal docs: https://vectorizedio.atlassian.net/wiki/spaces/CORE/pages/170491930/Features+Cluster+Logical+Versioning

versioned_docs/version-23.2/get-started/architecture.mdx
ztlpn: lgtm 👍

ztlpn (approved):

JoyceFee1 (approved):
Lgtm

versioned_docs/version-23.2/get-started/architecture.mdx
micheleRP: @mattschumpert Could you please confirm if there's anything in that [wiki ](https://vectorizedio.atlassian.net/wiki/spaces/CORE/pages/170491930/Features+Cluster+Logical+Versioning)that we should include in our doc? I think if it belongs anywhere, it should be the release notes, but please let me know if you think customers need some high-level explanation in doc of feature flags in Redpanda and how they work.

Feediver1 (approved):

versioned_docs/version-23.2/get-started/architecture.mdx
```diff
@@ -48,6 +48,12 @@ As long as there is a timing inequality between heartbeat time, election timeout
...
 
+## Controller partition and snapshots
+
+Redpanda stores metadata update commands (such as creating and deleting topics or users) in a system partition called the controller partition. Controller snapshots save the current cluster metadata state to disk, so startup is fast. Restoring from this snapshot only requires the time proportional to the number of metadata objects. A new snapshot is created after each controller command is added, or, with rapid updates, after a set period of time (default is 60 seconds).
```
mattschumpert, line 53: @micheleRP 
'time proportional to the number of metadata objects' is confusing.   

how and why is it faster than the alternative?

also we don't talk at all about how the manifest is compressed .

versioned_docs/version-23.2/get-started/architecture.mdx
```diff
@@ -48,6 +48,12 @@ As long as there is a timing inequality between heartbeat time, election timeout
...
+
+Controller snapshots are enabled by default in new clusters and are controlled by the `controller_snapshots` feature flag. Each broker has a snapshot file stored in the controller log directory; for example, `/var/lib/redpanda/data/redpanda/controller/0_0/snapshot`. The controller partition is replicated by a Raft group that includes all cluster brokers, and the controller snapshot is the Raft snapshot for this group. Snapshots are hydrated when a broker joins the cluster or restarts.
```
mattschumpert, line 55: @micheleRP enable the feature flag how?   Feature flags are normally not something users have to enable directly. 

@jcsp  for the beta, shouldn't we explain to users how to use the Admin API to turn on snapshots here (since we want to get user coverage during this time).  

For the GA version we shouldn't be instructing them to use a feature flag, as this should be a cluster config setting enabled by default

mattschumpert (approved):

versioned_docs/version-23.2/get-started/architecture.mdx
jcsp: >@jcsp for the beta, shouldn't we explain to users how to use the Admin API to turn on snapshots here (since we want to get user coverage during this time).

This would only be relevant if beta testers were upgrading existing systems to the beta.  In that edge case, they can be instructed directly, I don't think it needs to go in the public docs.

Documenting feature flags more generally and providing nice RPK command is perhaps something for the future, although I'm not sure you want that -- as you're saying it should become a config property rather than a feature flag here.  My advice would be:
- Don't document curl commands for feature flags -- at the point we want to encourage users to use them, create rpk commands.  This will probably be relevant the next time we add some tech preview feature, perhaps wasm.
- For cloud clusters, users will get controller snapshots enabled by us at some future time.
- For self hosted clusters, we can enable controller snapshots for upgrade clusters by default in some future release, rather than trying to coordinate enabling it everywhere (of course with the edge case of support cases where a cluster has an oversized log)

versioned_docs/version-23.2/get-started/architecture.mdx
micheleRP: @mattschumpert: I reworded as "Restoring from a snapshot only requires replaying the updates that have occurred since the last snapshot."
And the manifest info is in a different PR: https://github.com/redpanda-data/documentation/pull/1685

versioned_docs/version-23.2/get-started/architecture.mdx
micheleRP: @mattschumpert, @jcsp: I just rearranged this content a bit. [Please review](https://deploy-preview-1658--redpanda-documentation.netlify.app/docs/beta/get-started/architecture/#controller-partition-and-snapshots). Do you have any other changes to this for 23.2 (i.e., for now, do you want to remove "...and are controlled by the `controller_snapshots` feature flag")?

mattschumpert: @micheleRP . @jcsp  the fact that its not enabled by default /when its enabled needs some explanation, even if we want it to be 'contact us to enable' for beta.  @jcsp what do you think ? I understand we (Redpanda) will roll through cloud clusters using the Admin API  to test it out in stages but for the beta maybe it's just  'contact us', so e.g. Lacework can test it out.

@jcsp wrt GA, even if its off by default for now, for upgraded clusters I really think it should be a cluster config flag at 23.2 GA so customers can enable it at will without using our internal feature flag mechanism (even selectively at the direction of CS after a period of time).  This ought to be templat-able cluster configuration like any other not a special RPK command IMO and something e.g. a Lacework can refer back to to easily see the state of which clusters have it on or off.

versioned_docs/version-23.2/get-started/architecture.mdx
ztlpn: > Restoring from a snapshot only requires replaying the updates that have occurred since the last snapshot.

This is not true :) the original wording is correct. If it is confusing, we can add an example (e.g. a partition that is moved a few times).

micheleRP: Thanks @ztlpn! How about this: _Controller snapshots save the current cluster metadata state to disk, so startup is fast. For example, with a partition that has moved several times, a snapshot can restore the latest state without replaying every move command._

versioned_docs/version-23.2/get-started/architecture.mdx
micheleRP: Updated to this, for now: _Controller snapshots are enabled by default in new clusters. To enable controller snapshots in existing clusters, contact [Redpanda Support](https://support.redpanda.com/hc/en-us?__hstc=76629258.0157e0b034a780996d0c28de97dba0c6.1685652329951.1687298765119.1687306852922.3&__hssc=76629258.1.1687306852922&__hsfp=4199685904)._

versioned_docs/version-23.2/get-started/architecture.mdx
ztlpn: Sounds good!

versioned_docs/version-23.2/get-started/architecture.mdx
micheleRP: @mattschumpert please confirm if you want me to merge this now in Beta docs.

mattschumpert: @micheleRP I think we can merge this as is and direct folks to the feature flag as needed for the beta period.

@jcsp given the multi-release phased rollout of the feature, I think the GA should include the cluster config config flag, but IMO nothing else.  DO I need to file a separate ticket to ask for that cluster config toggle or can you just add it.  We had a similar situation with Server Side Schema Validation in reviewing the beta doc and Ben handled it (post-beta).

jcsp: Please reach out to the developers working on the feature with requests for changes, no need to route through me.

PR #1672, opened 2023-06-07
Add `Supported versions` field for properties introduced in 23.x

### LINKED ISSUE #1669: Document for each 23.x property the Redpanda version it was introduced in
### Describe the Issue

Request from https://redpandadata.slack.com/archives/C01LK3UK2K1/p1686008656170229:

> when was log_segment_ms introduced? Is that a 23.x -only setting? Lacework is trying to set this on 22.3.13 and getting an unknown setting error. Trying to figure out if there's something else we should be modifying.

> It would be nice if the docs had a note like "parameter introduced in XX.Y.Z"

shorgi:
Fixes #1669 

Review deadline: Tue Jun 13

Reviewers: in Files changed, see the properties that have `Supported versions` added. (Note: 23.2 Beta properties pages have also been updated with missing properties)


Feediver1 (approved):
lgtm

hcoyote (approved):

PR #1690, opened 2023-06-13
Switch owl-shop example to use yaml config.

voutilad:
Environment variables seem busted. Also adds a healthcheck for Redpanda as owlshop will fail fast and not retry to connect if the broker interface isn't ready.

### DIFF
--- a/docs/reference/docker-compose.mdx
+++ b/docs/reference/docker-compose.mdx
@@ -72,6 +72,12 @@ This `docker-compose.yml` file starts a single Redpanda broker, Redpanda Console
         - redpanda:/var/lib/redpanda/data
       networks:
         - redpanda_network
+      healthcheck:
+        test: ["CMD-SHELL", "rpk cluster health | grep -E 'Healthy:.+true' || exit 1"]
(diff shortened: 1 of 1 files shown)


larsenpanda:
I'm using a docker-compose.yml from april 3 and it works great. Josh Purcell also tried it and it works. Not sure what all changed and why. I've attached a zip of the yaml, apparently gh doesn't like straight yaml
[docker-compose.yml.zip](https://github.com/redpanda-data/documentation/files/11738434/docker-compose.yml.zip)
y.

voutilad:
Upstream owl-shop has [changes](https://github.com/cloudhut/owl-shop/commit/042112b75cd0533733785bddd9c15d3d6afcd0f8) it seems completely overhauling the configuration. Since they never published an arm64 image, building a new image from source requires modifying the config approach.

Maybe we should host a known-good version of the project so we can incorporate an arm64 image? It's all pure Go code and super annoying that there's only an amd64 version.

JakeSCahill (approved):
Nice, thanks @voutilad

PR #1536, opened 2023-04-21
Add Benchmark Redpanda as a Develop topic

shorgi:
Review deadline: ~~Mon May 1~~ Wed Jun 7

Fixes #1202 

Preview:
- New [Benchmark Redpanda topic](https://deploy-preview-1536--redpanda-documentation.netlify.app/docs/develop/benchmark/)

PR #1715, opened 2023-06-26
Fix regression for editable code

shorgi:
Fix a regression in editable code:
- In https://github.com/redpanda-data/documentation/pull/1536#discussion_r1177128927, within a draft PR, XML was excluded from editable code.
- Before that was merged, in https://github.com/redpanda-data/documentation/pull/1631/files#diff-3940d505894d7c25e65f3f81174db31b755777d8cf5e23267df184eddd29e361, a similar fix was implemented but it actually disabled all editable code.
- This PR changes the implementation back to the first one.


Deflaimun (approved):

JakeSCahill (approved):
Thanks 😊

Issue #1608, opened 2023-05-22
Document `rpk profile` and `rpk cloud auth`, `-X` flag new commands for v23.2


twmb:
### Describe the Issue

We are introducing a few new features in rpk in v23.2.
* a global `-X` flag to setting configuration
* a new command space, `rpk profile`
* a new command space, `rpk cloud auth`
* changes in behavior the existing command `rpk cloud login`

There are a lot of changes, so we should work tightly to figure out what all needs to be documented. I can try to expand on detail in this ticket, but there's a lot of detail, so it may be worth a quick sync first to break down how we'd like to attack this.

* Pull requests introducing the functionality: https://github.com/redpanda-data/redpanda/pull/9981 and https://github.com/redpanda-data/redpanda/pull/10528
* SMEs: @twmb, @r-vasquez 
* Channel: #discuss-devex
* Included in release v23.2

twmb:
* We have added `{edit,list,set}-defaults` to `rpk profile`
* We have added `prompt` -- which probably requires a bit of explanation

Three more related PRs:
* https://github.com/redpanda-data/redpanda/pull/11296
* https://github.com/redpanda-data/redpanda/pull/11353
* https://github.com/redpanda-data/redpanda/pull/11444

One more small PR is coming, I will update after it is merged

PR #1716, opened 2023-06-26
rpk: adding sasl flags to rpk redpanda admin commands

r-vasquez:
These were missing and now are being added in
redpanda/#11650: https://github.com/redpanda-data/redpanda/pull/11650


shorgi (approved):
#1608 will also update all rpk command reference pages

PR #1756, opened 2023-07-13
Add IDs to issue template

JakeSCahill:
Having IDs for each field will allow us to auto-populate the form using URL query parameters.

See https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#keys


Deflaimun (approved):
LGTM

PR #1764, opened 2023-07-14
docs: #1758 explain controller snapshots are off by default in upgrade…

### LINKED ISSUE #1758: Explain controller snapshots are off by default for upgraded clusters in 23.2
### Issue description

https://docs.redpanda.com/docs/beta/get-started/architecture/#optimized-platform-performance explains controller snapshots 

For new clusters its on by default but for upgraded clusters, 23.2.1 has it switched off.

This page and the upgrade page (https://docs.redpanda.com/docs/beta/manage/cluster-maintenance/rolling-upgrade/) should have a note that Controller Snapshots are disabled by default for upgraded clusters.

We should also explain how to enable controller snapshots with the config property. @VladLazar  is the SME for the configuration/behavior

### Link to Redpanda Slack conversation

_No response_

micheleRP:
resolves #1758 

preview:


micheleRP:
@VladLazar can you please confirm if we should document how to enable this? For some reason I thought only support should do this for 23.2, but if you want I can add to the doc this section: https://vectorizedio.atlassian.net/wiki/spaces/CORE/pages/382468112/Controller+snapshots#Enabling

mattschumpert (changes requested):
We need the explicit config to enable it not to say contact support

mattschumpert:
@micheleRP we can use the REST API for now (beta) to explain this is needed on clusters upgraded to 23.2 from 23.1 and earlier.

mattschumpert (approved):

ztlpn (approved):

shorgi (approved):

PR #1767, opened 2023-07-18
docs: #1475 add 2 settings to mm2 docs

### LINKED ISSUE #1475: Add 2 settings to mm2 docs
### Describe the Issue

ACL's and Consumer Offsets sync is not well documented (and from what i can tell, consumer offset sync is not doc'd at all) on the official mm2 docs, we can benefit from that in SEO and of course helping our customers/users.

For ACL's, if the source cluster doesn't have ACL's applied mm2 will fail if you don't tell it not to try to sync them.

For consumer offsets, the default is to not sync them and there's no good documentation on that configuration.

The two settings are:

`sync.topic.acls.enabled = false` to not sync acl's

and

`sync.group.offsets.enabled = true` to sync consumer offsets

micheleRP:
resolves #1475 

preview: https://deploy-preview-1767--redpanda-documentation.netlify.app/docs/deploy/deployment-option/cloud/managed-connectors/create-mmaker-source-connector/


micheleRP:
@larsenpanda looks like `sync.group.offsets.enabled` defaults to `true` now, so this PR just adds a tip about changing the default for `sync.topic.acls.enabled` if the source doesn't have ACLs, and it does some cleanup to match the UI. Does this need anything else?

larsenpanda (approved):
lgtm!

Deflaimun (approved):

PR #1771, opened 2023-07-19
tls: Update preferred ciphers

StephanDollberg:
We did recently change our preferred ciphers to optimize them for performance.

Update the docs respectively.

Note that the TLS v1.3 preferred was incorrect but is now correct so no change needed there.

This change became active in v23.1.12 and v23.2+. Is this something we care about here? Do I somehow have to version the docs (I see some docs have a versions). It's just a tiny change so I am not sure it matters much from a docs perspective.


Feediver1 (approved):
lgtm

PR #1822, opened 2023-08-01
Improve description of raft_learner_recovery_rate

### LINKED ISSUE #1403: Add more detail to raft_learner_recover_rate definition
Ah, got it. This is a little vague to me (from the reference docs).

&gt;  Raft learner recovery rate limit. Throttles the rate of data communicated to nodes (learners) that need to catch up to leaders.
From <https://docs.redpanda.com/docs/reference/cluster-properties/#raft_learner_recovery_rate>

Probably worth adding something like:

"This is a rate limit placed on the node sending data to the recovering node. Each sending node will be limited to this rate. The recovering node will accept data as fast as possible according to the combined limits of each healthy nodes in the cluster. For example, if two nodes are sending data to the recovering node, and the raft_learner_recover_rate limit is 100MBytes/sec, then the recovering node will recover at a rate of 200MBytes/sec"

[Slack Message](https://redpandadata.slack.com/archives/C03KAHQ776C/p1679074452219469?thread_ts=1679073243.950889&cid=C03KAHQ776C)

shorgi:
Fixes #1403

### DIFF
--- a/docs/reference/cluster-properties.mdx
+++ b/docs/reference/cluster-properties.mdx
@@ -1079,7 +1079,9 @@ The cluster metrics of the metrics reporter are different from [monitoring metri
 
 ### raft_learner_recovery_rate
 
-Raft learner recovery rate limit. Throttles the rate of data communicated to nodes (learners) that need to catch up to leaders.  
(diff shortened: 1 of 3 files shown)


hcoyote (approved):
LGTM

Feediver1 (approved):
lgtm

PR #1827, opened 2023-08-01
Use base-uri variable for curl requests

### LINKED ISSUE #1614: Panda Proxy - Update base URL that is required for future calls
### Describe the Issue

https://docs.redpanda.com/docs/develop/http-proxy/

Feedback from CS team:  The base URL is needed for future calls, however, the doc only shows the doc a local host. Please update the docs to include the base URL in place of the local host.

This link https://docs.redpanda.com/docs/develop/http-proxy/#consume-topic-events-with-consumers should probably be called "create a consumer" or something...

However, the bigger issue is that the call gives back a baseURI, which we note in the docs, but we need to use that baseURI in the calls below, such as "subscribe to the topic" / "get data", etc

Any questions? Contact Paul Wilkinson (CS team)

shorgi:
Fixes #1614

### DIFF
--- a/docs/develop/http-proxy.mdx
+++ b/docs/develop/http-proxy.mdx
@@ -345,7 +345,7 @@ Expected output:
 [{"topic":"test_topic","key":null,"value":"Redpanda","partition":0,"offset":0}]
 ```
 
-### Consume topic events with consumers
+### Create a consumer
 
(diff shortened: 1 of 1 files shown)


Deflaimun (approved):

Deflaimun:
The change is good but somehow the placeholder is not editable. Checking later...

Feediver1 (approved):
lgtm

PR #1831, opened 2023-08-02
Add Configure Offset Retention

### LINKED ISSUE #1555: Document OffsetDelete Kafka API (23.2)
https://github.com/redpanda-data/redpanda/pull/8669

- Document its usage in terms of culling consumer offsets (more advanced/manual approach than the auto expiry feature added in 23.1)
- Remove the gap reference from our Kafka compat docs page

shorgi:
Fixes 
- #1555

### DIFF
--- a/docs/manage/cluster-maintenance/disk-utilization.mdx
+++ b/docs/manage/cluster-maintenance/disk-utilization.mdx
@@ -108,7 +108,7 @@ To set retention size for a single topic, use `retention.bytes`, which overrides
 
 Redpanda supports consumer group offset retention through both periodic offset expiration and the Kafka OffsetDelete API. 
 
-For periodic offset expiration, you set the retention duration of consumer group offsets and the check period, and Redpanda identifies offsets that are expired and removes them to reclaim storage. For a consumer group, the retention timeout starts from when the group becomes empty by losing all its consumers. For a standalone consumer, the retention timeout starts from the time of the last commit. Once elapsed, an offset is considered to be expired and is discarded.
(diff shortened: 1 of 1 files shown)


mattschumpert (approved):

docs/manage/cluster-maintenance/disk-utilization.mdx
```diff
@@ -104,6 +104,20 @@ To set retention size for a single topic, use `retention.bytes`, which overrides
   rpk cluster config set retention_bytes <retention_size>
   ```
 
+## Configure offset retention
+
+Redpanda supports consumer group offset retention through both periodic offset expiration and the Kafka OffsetDelete API. 
+
+For periodic offset expiration, you set the retention duration of consumer group offsets and the check period, and Redpanda identifies offsets that are expired and removes them to reclaim storage. For a consumer group, the retention timeout starts from when the group becomes empty by losing all its consumers. For a standalone consumer, the retention timeout starts from the time of the last commit. Once elapsed, an offset is considered to be expired and is discarded.
```
Feediver1, line 111:
```suggestion
For periodic offset expiration, set the retention duration of consumer group offsets and the check period. Redpanda identifies offsets that are expired and removes them to reclaim storage. For a consumer group, the retention timeout starts from when the group becomes empty as a consequence of losing all its consumers. For a standalone consumer, the retention timeout starts from the time of the last commit. Once elapsed, an offset is considered to be expired and is discarded.
```

Feediver1 (approved):

PR #1839, opened 2023-08-04
docs: #1815 multiple RP instances can't be on same host

### LINKED ISSUE #1815: Document requirement that multiple RP instances are not on same physical host
----

_This action was created from [Incident 323](https://app.incident.io/redpandadata/incidents/323) by Todd Farmer, using [incident.io](https://app.incident.io) 🔥_

micheleRP:
resolves #1815 

preview:

### DIFF
--- a/docs/deploy/deployment-option/self-hosted/kubernetes/kubernetes-cluster-requirements.mdx
+++ b/docs/deploy/deployment-option/self-hosted/kubernetes/kubernetes-cluster-requirements.mdx
@@ -56,6 +56,10 @@ The Helm chart configures `podAntiAffinity` rules to make sure that each Redpand
 
 Redpanda brokers must store their data on disk. As a result, each Pod that runs a Redpanda broker must have its own storage volume.
 
+:::note
+Do not colocate multiple Redpanda instances on a single host in Kubernetes.
+:::
+
 For production, follow these requirements and recommendations:
 
 - Mount an XFS or ext4 file system on any storage volumes that host the data directory of Redpanda (`/var/lib/redpanda/data`) or the Tiered Storage cache. XFS is highly recommended. NFS is not supported.


micheleRP:
Hi @andrwng, @WesWWagner, @hcoyote: Please review this note and confirm if we should include anything else. Should we say that it's specific to GCP (not a problem on AWS)?

hcoyote:
> Hi @andrwng, @WesWWagner, @hcoyote: Please review this note and confirm if we should include anything else. Should we say that it's specific to GCP (not a problem on AWS)?

I'm not sure this is the right message. 

In this particular case, the user placed every single redpanda pod on the same k8s node, which is ultimately what was causing the problem. There's definitely use cases out there where a user could have a single shared k8s cluster and multiple sets of small redpanda pods comprising smaller cluster configurations where having multiple redpanda pods on the same node would be ok IMO (e.g., they're properly restricting memory and cpu).

micheleRP:
> > Hi @andrwng, @WesWWagner, @hcoyote: Please review this note and confirm if we should include anything else. Should we say that it's specific to GCP (not a problem on AWS)?
> 
> I'm not sure this is the right message.
> 
> In this particular case, the user placed every single redpanda pod on the same k8s node, which is ultimately what was causing the problem. There's definitely use cases out there where a user could have a single shared k8s cluster and multiple sets of small redpanda pods comprising smaller cluster configurations where having multiple redpanda pods on the same node would be ok IMO (e.g., they're properly restricting memory and cpu).

Thanks @hcoyote. Would it be sufficient to say:
 Do not place all your Redpanda instances on a single host in Kubernetes.

And it is OK as a plain note, or should this be emphasized visually as an important note?

Feediver1 (approved):
lgtm

WesWWagner (approved):
While previous commenters did note that it is possible to safely run mutliple redpanda clusters on the same nodes if a large list of preconditions were considered, I believe this note is straight forward enough to stop issues from occurring and we can deal with the "why not" requests of people who want to run multiple clusters in Customer Success and presales.

WesWWagner:
> Hi @andrwng, @WesWWagner, @hcoyote: Please review this note and confirm if we should include anything else. Should we say that it's specific to GCP (not a problem on AWS)?

Just to add for history here, if you are using ephemeral storage and not properly configuring rack awareness to nodes you could lose data from the loss of a single piece of hardware. There are too many things that need to go right to co home multiple nodes on a single k8s node, and then why bother? You could just run a single node with more CPUs.

micheleRP:
> While previous commenters did note that it is possible to safely run mutliple redpanda clusters on the same nodes if a large list of preconditions were considered, I believe this note is straight forward enough to stop issues from occurring and we can deal with the "why not" requests of people who want to run multiple clusters in Customer Success and presales.

Thank you, @WesWWagner!

PR #1844, opened 2023-08-08
Add note explaining offset and timestamp behavior in mm2 doc

### LINKED ISSUE #1802: document timestamp and offset behaviour with MM2
### Issue description

We had frequent questions on behaviour of offsets and timestamps during recent MM2 work...  so could we please document is.

- Timestamps
The events written on the target topic will use the timestamp that was set on the source event.
e.g if on the source the event had a timestamp which was  "2023-05-22 17:00" that would also be the timestamp on the target event 

- Offsets
Offsets are not guaranteed to match between the source and target. 
For example if data-retention deletes have occurred  on the source topic the earliest offset could be for example #5000
but when that event is created on the source topic the offset for that event will be #0 

https://docs.redpanda.com/docs/deploy/deployment-option/cloud/managed-connectors/create-mmaker-source-connector/

### Link to Redpanda Slack conversation

_No response_

Feediver1:
Fixes #1802

### DIFF
--- a/docs/deploy/deployment-option/cloud/managed-connectors/create-mmaker-source-connector.mdx
+++ b/docs/deploy/deployment-option/cloud/managed-connectors/create-mmaker-source-connector.mdx
@@ -31,6 +31,17 @@ You can also use it to:
 - Only ACLs for topics matching the connector configuration are copied (write ACLs are not copied).
 - All permissions ACLs are downgraded to read-only.
 
+:::note
+Offsets are not guaranteed to match between the source and target. For example,
+if data-retention deletes occur on the source topic and the earliest
+offset is `#5000`, then when that event is created on
+the source topic the offset for that event will be `#0`.
+
+Events written on the target topic use the timestamp that was set on the source event.
+For example, if the source event has a timestamp `2023-05-22 17:00`, then this
+would also be the timestamp on the target event.
+:::
+
 ## Create a MirrorMaker2 Source connector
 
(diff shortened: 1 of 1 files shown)


Feediver1:
Looking at this now, wondering if this even needs to be a note @bochenekmartin and @tomasz-sadura . Can I just add as another bullet in Limitations? thx

jason-da-redpanda (approved):

docs/deploy/deployment-option/cloud/managed-connectors/create-mmaker-source-connector.mdx
```diff
@@ -31,6 +31,17 @@ You can also use it to:
 - Only ACLs for topics matching the connector configuration are copied (write ACLs are not copied).
 - All permissions ACLs are downgraded to read-only.
 
+:::note
+Offsets are not guaranteed to match between the source and target. For example,
+if data-retention deletes occur on the source topic and the earliest
+offset is `#5000`, then when that event is created on
+the source topic the offset for that event will be `#0`.
```
tomasz-sadura, line 38:
```suggestion
the target topic the offset for that event will be `#0`.
```

tomasz-sadura (approved):

tomasz-sadura:
I wouldn't treat it as limitation, it is just how it works.

jason-da-redpanda (approved):
oops + thanks :-)

micheleRP (approved):
lgtm

PR #1849, opened 2023-08-10
Remove max message size comment

vuldin:
Redpanda now has a max message size unlike in previous versions.

### DIFF
--- a/docs/manage/data-migration.mdx
+++ b/docs/manage/data-migration.mdx
@@ -163,7 +163,7 @@ To stop the MirrorMaker 2 process, use `top` to find its process ID, and then ru
 
 When replicating larger message sizes with MirrorMaker 2 on the target cluster, you may get blocked with an error:
 ```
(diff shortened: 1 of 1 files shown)


Deflaimun (approved):

docs/manage/data-migration.mdx
```diff
@@ -161,10 +161,7 @@ To stop the MirrorMaker 2 process, use `top` to find its process ID, and then ru
 
 ## Message size
 
-Apache Kafka® limits you to 1MB max message size. Redpanda doesn't impose such limits. 
-
-However, when replicating those larger message sizes with MirrorMaker 2 on the target cluster, you can get blocked with an error 
-
+When replicating larger message sizes with MirrorMaker 2 on the target cluster, you may get blocked with an error:
 ```
 org.apache.kafka.common.errors.RecordTooLargeException: The message is xxxx bytes when serialized which is larger than 1048576, which is the value of the max.request.size configuration
```
Feediver1, line 166:
```suggestion
org.apache.kafka.common.errors.RecordTooLargeException: The message is xxxx bytes when serialized which is larger than 1048576, which is the value of the max.request.size configuration.
```

PR #1857, opened 2023-08-15
Fix rpk commands in Kubernetes docs

JakeSCahill:
- Added namespace placeholder to broker addresses so that they are updated dynamically.
- Updated `--brokers`, `--admin-*`, and `--api-urls` to the new `-X` flags.


Feediver1 (approved):
lgtm

PR #1897, opened 2023-08-23
Label cloud pages with attributes

JakeSCahill:
Allows us to hide the version selector on any pages marked with `:page-cloud: true`

Relies on https://github.com/redpanda-data/docs-ui/pull/95


Deflaimun (approved):

Feediver1:
In the preview, I still see cloud content is versioned. Would the removal be handled in a separate PR @JakeSCahill ? We don't want users to have to opt out of versioning when in the context of cloud.

JakeSCahill:
@Feediver1 yeah, the UI handles that and we just published the update.

